PR #129, opened 2016-08-12
Added SingletonPerThread provider

sirkonst:
Very useful for me :-)

### DIFF
--- a/dependency_injector/providers/creational.py
+++ b/dependency_injector/providers/creational.py
@@ -1,6 +1,7 @@
 """Dependency injector creational providers."""
 
 import six
+from threading import local
 
 from dependency_injector.providers.callable import Callable
 from dependency_injector.providers.base import _parse_keyword_injections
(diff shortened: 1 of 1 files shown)


rmk135:
Hello,

Just wanted to say "Thanks". I will take a look a bit latter.

Roman

rmk135:
Hello Konstantin,

I've reworked a bit your idea and merged it in terms of PR #132. Also I've added a special section in documentation about using singletons in multi-threading applications.

I productized new version with number 2.1.0 (http://python-dependency-injector.ets-labs.org/en/stable/main/changelog.html) and it's been available on PyPi (https://pypi.python.org/pypi/dependency_injector/2.1.0), so you're welcome to upgrade.

PS: Thanks a lot for your contribution, I have added you in a list of contributors (https://github.com/ets-labs/python-dependency-injector/blob/master/CONTRIBUTORS.rst). Hope you are fine with that ;)

Thanks,
Roman

sirkonst:
Thanks! Great library!

rmk135:
Thank you 😊

PR #133, opened 2016-09-06
Normalize package names by PEP-503

sirkonst:
See https://www.python.org/dev/peps/pep-0503/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ class PublishCommand(Command):
         os.system('git push --tags')
 
 
-setup(name='dependency_injector',
+setup(name='dependency-injector',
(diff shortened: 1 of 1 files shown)


rmk135:
Hello Konstantin,

Thanks for this, once more :) I will take a detail look and merge it a bit later, cause I really want to check that everything (links, CI, etc...) would work as expected.

Thanks,
Roman

rmk135:
Thanks :)

rmk135:
Change has been productized in version 2.1.1 - available on PyPI https://pypi.python.org/pypi/dependency_injector

sirkonst:
Great!

PR #227, opened 2019-08-17
Minor English nitpicking.

supakeen:
Hey there, I was feeling nitpicky today so I went over your README and made several small changes to make the English flow a bit better.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,19 +3,20 @@ Dependency Injector - Dependency injection microframework for Python
 ====================================================================
 
 *Dependency Injector* is a dependency injection microframework for Python. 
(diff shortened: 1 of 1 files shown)


rmk135:
@supakeen , thanks a lot, that's just great!

Definitely merging this! Just in case you would have some time and same nitpicking mood -  you're very welcome to do the same for `/docs`.

Thanks,
Roman

rmk135:
This change is productized in `3.14.9`, available in master and on PyPi. @supakeen, I've also added you to the list of contributors - https://github.com/ets-labs/python-dependency-injector/blob/master/CONTRIBUTORS.rst.

Thanks again 👍

PR #252, opened 2020-06-16
Update singleton_thread_locals.py

RobinsonMa:
fix

### DIFF
--- a/examples/providers/singleton_thread_locals.py
+++ b/examples/providers/singleton_thread_locals.py
@@ -1,7 +1,7 @@
 """`ThreadLocalSingleton` providers example."""
 
 import threading
-from queue import Queue
+import queue
 
 import dependency_injector.providers as providers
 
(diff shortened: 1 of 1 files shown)


examples/providers/singleton_thread_locals.py
```diff
@@ -15,7 +15,7 @@ def example(example_object, queue):
 thread_local_object = providers.ThreadLocalSingleton(object)
 
 # Create singleton provider for thread-safe queue:
-queue = providers.Singleton(Queue.Queue)
+queue = providers.ThreadSafeSingleton(queue.Queue)
```
rmk135, line 18: Please, rename `queue` to `queue_factory`. As of now, `queue` variable overrides the name of the imported module. It's a bad practice. Also `queue_factory` name will be consistent with `thread_factory` at line 26.

rmk135:
@RobinsonMa , thank you for contribution. It will be on production today.

RobinsonMa:
@rmk135 ,It's my pleasure

rmk135:
@RobinsonMa , the code is on production http://python-dependency-injector.ets-labs.org/providers/singleton.html#singleton-providers-and-multi-threading

I've added you to CONTRIBUTORS file. It's distributed with an every copy - https://github.com/ets-labs/python-dependency-injector/blob/master/CONTRIBUTORS.rst

Changelog is also updated. See version `3.16.1` at http://python-dependency-injector.ets-labs.org/main/changelog.html

Roman

PR #322, opened 2020-11-10
Fix race in ThreadSafeSingleton

rda-dev:
The `self.__storage` attribute may be overwritten by the `ThreadSafeSingleton.reset()` method and thus will return a None from the `_provide(...)` method.

Additionally, we should place a return statement inside the lock to prevent reading the shared variable. A more efficient solution is 'double-checked locking optimization.'

### DIFF
--- a/src/dependency_injector/providers.pyx
+++ b/src/dependency_injector/providers.pyx
@@ -2178,15 +2178,19 @@ cdef class ThreadSafeSingleton(BaseSingleton):
 
         :rtype: None
         """
-        self.__storage = None
+        with self.__storage_lock:
+            self.__storage = None
 
     cpdef object _provide(self, tuple args, dict kwargs):
         """Return single instance."""
-        with self.__storage_lock:
-            if self.__storage is None:
-                self.__storage = __factory_call(self.__instantiator,
-                                                args, kwargs)
-        return self.__storage
+        storage = self.__storage
+        if storage is None:
+            with self.__storage_lock:
+                if self.__storage is None:
+                    self.__storage = __factory_call(self.__instantiator,
+                                                    args, kwargs)
+                storage = self.__storage
+        return storage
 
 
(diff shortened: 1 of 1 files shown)


rmk135:
Hey @rda-dev ,

Thanks for the PR. This looks very good. A cosmetic question before merging - do we need ``storage`` temp variable? Does it have any special meaning?

If it's not required, can you, please, remove it in favour of ```self.__storage``?

rda-dev:
Hi @rmk135 ,

Yes, the first `storage` local variable assignment captures an attribute value and allows not to use lock if the variable is already initialized. The second `storage` assignment may be replaced with `return self.__storage` statement, but I think it will be clearer to leave a single return from the method and reuse the storage variable here. Isn't it?

rmk135:
I like to avoid temporary variables when possible. Here is what I meant:

```python
        if self.__storage is None:
            with self.__storage_lock:
                if self.__storage is None:
                    self.__storage = __factory_call(self.__instantiator,
                                                    args, kwargs)
        return self.__storage
```

It will work the same way, right?

rda-dev:
No, it's not the same. In your example, when you release the lock, the parallel thread may set the `__storage` attribute to None (with `reset()` method) and you will return None instead of an initialized object.

rmk135:
Heh, that's tricky. I got the difference finally. Didn't consider ``reset()`` effect eventually. Thanks for explanation. Merging.

rmk135:
I will cythonize the code and make a release in a couple of hours.

rmk135:
@rda-dev I have released fix in version `4.3.7`. See changelog https://python-dependency-injector.ets-labs.org/main/changelog.html.

I have also added you to the list of contributors https://github.com/ets-labs/python-dependency-injector/blob/master/CONTRIBUTORS.rst. It is distributed with every copy of Dependency Injector.

Thank you again for the fix and for explaining how it works.

Best,
Roman

PR #440, opened 2021-03-29
Fix provider documentation

sonthonaxrk:
Make example more realistic.  This would break if you're running flask with multiple threads.


rmk135:
Thanks for the contribution. I published the fix in version `4.31.2`. I've also added you to the list of contributors https://github.com/ets-labs/python-dependency-injector/commit/664a6ca5d947d2c6e51ec697d908b7d0bec0735c

PR #442, opened 2021-03-29
ContextLocalSingleton Provider Class

sonthonaxrk:
This is just a preliminary example of how I'd switch to using ContextVars.  The advantage of ContextVars is that they actually work when using asyncio.  I think it might be worth having a `ContextLocalSingleton` and a `ThreadLocalSingleton`.  Most people will need a `ContextLocalSingleton`.

Things I need to do here.  Please add anything I'm missing :)

- [x] Add tests
- [x] Fix the pipeline
- [x] Add the backported `contextvars` library to the dependencies so we can support earlier Python 3 versions.
- [x] Handle Python 2.  I think it might make sense to tie the class into the 'async enabled' flag you have.

Thanks for all your work on this library.


sonthonaxrk:
Build is red because of coveralls failing to upload the results.

```
coveralls run-test: commands[2] | coverage report --rcfile=./.coveragerc
Name                                      Stmts   Miss  Cover
-------------------------------------------------------------
src/dependency_injector/__init__.py           2      0   100%
src/dependency_injector/containers.pyx      343      9    97%
src/dependency_injector/errors.py             2      0   100%
src/dependency_injector/ext/__init__.py       0      0   100%
src/dependency_injector/ext/aiohttp.py       21      0   100%
src/dependency_injector/ext/flask.py         38      2    95%
src/dependency_injector/providers.pxd       173     21    88%
src/dependency_injector/providers.pyx      1731    106    94%
src/dependency_injector/resources.py         25      6    76%
src/dependency_injector/schema.py           147     20    86%
src/dependency_injector/wiring.py           518     26    95%
-------------------------------------------------------------
TOTAL                                      3000    190    94%
```

requirements.txt
```diff
@@ -1 +1,2 @@
 six>=1.7.0,<=1.15.0
+contextvars~=2.4; python_version >= "3.0" and python_version < "3.7"
```
sonthonaxrk, line 2: As a note, this library is stable, and has been on version 2.4 for the last 4 years.

rmk135: Got it, that's good. I worry about adding the dependency cause it could cause a dependencies resolving problem with other libraries. What if we remove it from the main dependencies and modify the guard clause in the provider to something like this:

```
        if not contextvars:
            raise RuntimeError(
                'Contextvars library not found. This provider '
                'requires Python 3.7 or a backport of contextvars. '
                'To install a backport run "pip install contextvars".'
            )
```

How does it look to you?

src/dependency_injector/providers.pyx
```diff
@@ -2928,6 +2934,89 @@ cdef class ThreadLocalSingleton(BaseSingleton):
...
+
+        super(ContextLocalSingleton, self).__init__(provides, *args, **kwargs)
+        self.__storage = contextvars.ContextVar('__storage', default=self._none)
+
+    def reset(self):
+        """Reset cached instance, if any.
+
+        :rtype: None
+        """
+
```
rmk135, line 2977: Could you please remove this empty line? I usually try to avoid them.

src/dependency_injector/providers.pyx
```diff
@@ -4645,4 +4734,4 @@ cpdef str _class_qualname(object instance):
     name = getattr(instance.__class__, '__qualname__', None)
     if not name:
         name = '.'.join((instance.__class__.__module__, instance.__class__.__name__))
-    return name
\ No newline at end of file
+    return name
```
rmk135, line 4737: 👍

tests/unit/providers/__init__.py
```diff
@@ -1 +1,435 @@
 """Dependency injector providers unit tests."""
+import sys
```
rmk135, line 2: Not sure why did you move the tests here. Could we leave them on their place and just create a new module for contextvars singleton, e.g. `test_singletons_py3.py`?

sonthonaxrk: I think if I import `test_singletons_py2_py3.py` from `test_singletons_py3.py` the python2 tests will get run.

I could be wrong though.

I could move what I added to `__init__.py` to `singleton_test_common_utils.py` and import it from both files.

src/dependency_injector/providers.pyx
sonthonaxrk: Vim sometimes strips new lines at the end of files.

src/dependency_injector/providers.pyx
sonthonaxrk: Sure

requirements.txt
sonthonaxrk: Sure thing, no objections.  I'm not really concerned about people on older Python versions.

However we could probably get rid of the `~=2.4` version constraint and still be fine.  Considering just how old that library is.

sonthonaxrk:
@rmk135 

I've made the amends and rebased onto dev.

Makefile
```diff
@@ -41,6 +41,7 @@ docs-live:
 
 install: uninstall clean cythonize
 	pip install -ve .
+	pip install contextvars
```
sonthonaxrk, line 44: Since the older python3 tests depend on this.

sonthonaxrk:
Fixed the pipeline.

sonthonaxrk:
Hey @rmk135 are you still interested in merging this?

rmk135:
> Hey @rmk135 are you still interested in merging this?

Hey @sonthonaxrk , I do. Merging it now :)

billcrook:
Just a heads up that there is no documentation of this provider in the main docs

rmk135:
@billcrook , thanks for the heads up. It's still in my backlog, will prioritize it!

PR #490, opened 2021-08-16
Remove generic meta class from resource and async resource classes

rmk135:


src/dependency_injector/resources.py
```diff
@@ -31,7 +18,7 @@ def shutdown(self, resource: T) -> None:
         ...
 
 
-class AsyncResource(Generic[T], metaclass=ResourceMeta):
+class AsyncResource(Generic[T]):
```
EdwardBlair, line 21: Does this still not need to inherit from abc.ABC?

src/dependency_injector/resources.py
```diff
@@ -1,26 +1,13 @@
...
 
 
-class ResourceMeta(GenericMeta, abc.ABCMeta):
-    def __getitem__(cls, item):
-        # Spike for Python 3.6
-        return cls(item)
-
-
-class Resource(Generic[T], metaclass=ResourceMeta):
+class Resource(Generic[T]):
```
EdwardBlair, line 10: Does this still not need to inherit from abc.ABC?

src/dependency_injector/resources.py
rmk135: Hm... It probably needs to. I didn't have a test for unimplemented abc methods. I'll revise this later today. Thanks!

rmk135: Returned ``abc.ABCMeta`` inheritance and added tests in https://github.com/ets-labs/python-dependency-injector/commit/1163ac59d4443c1cfbe2c4391752c83512086b99. I should be more watchful. Thanks for pointing out the issue.

Issue #509, opened 2021-09-14
Better error handling


gen-xu:
Currently if the callable provider fails due to missing arguments or other errors, it may be hard to find out which guy is missing arguments if one provider is being referred many times.
```pytb
src/dependency_injector/providers.pxd:422: in dependency_injector.providers.__provide_keyword_args
    ???
src/dependency_injector/providers.pxd:345: in dependency_injector.providers.__get_value
    ???
src/dependency_injector/providers.pyx:195: in dependency_injector.providers.Provider.__call__
    ???
src/dependency_injector/providers.pyx:197: in dependency_injector.providers.Provider.__call__
    ???
src/dependency_injector/providers.pyx:2755: in dependency_injector.providers.Singleton._provide
    ???
src/dependency_injector/providers.pxd:620: in dependency_injector.providers.__factory_call
    ???
src/dependency_injector/providers.pxd:606: in dependency_injector.providers.__callable_call
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   TypeError: __init__() missing 1 required positional argument: 'config'

src/dependency_injector/providers.pxd:579: TypeError
```
it may be hard to find out which `__init__`, consider re-raise the exception with the function reference

rmk135:
Hi @gen-xu. Yes, I agree that it turns to be a problem. Thanks for the PR. I'll look at it later today.

rmk135:
@gen-xu , commented on the PR.

attentive:
Nothing much to add, just having my first experiences with dependency-injector after using lots of other DI stacks, I have about 20 items in my container and am currently having trouble locating the place __init__() is failing. 

Library seems great otherwise 👏

thatguysimon:
Hey @rmk135, error handling still seems to be an issue. Any updates on this? I see there are a couple of PRs attempting to solve this.

dharmagetic:
@rmk135 I also request this issue

PR #510, opened 2021-09-14
Better error handling: Re-raise with function reference on error

gen-xu:
This PR addresses https://github.com/ets-labs/python-dependency-injector/issues/509

PR #528, opened 2021-11-01
Better error handling #509

whysage:
Add providers log on error.

### DIFF
--- a/src/dependency_injector/providers.pyx
+++ b/src/dependency_injector/providers.pyx
@@ -204,7 +204,11 @@ cdef class Provider(object):
         if self.__last_overriding is not None:
             result = self.__last_overriding(*args, **kwargs)
         else:
-            result = self._provide(args, kwargs)
+            try:
+                result = self._provide(args, kwargs)
+            except Exception as exc:
+                print(self)
+                raise exc
 
         if self.is_async_mode_disabled():
             return result
--- a/src/dependency_injector/providers.pyx
+++ b/src/dependency_injector/providers.pyx
@@ -6,6 +6,7 @@ import copy
 import errno
 import functools
 import inspect
+import logging
 import os
 import re
 import sys
@@ -207,7 +208,7 @@ cdef class Provider(object):
             try:
                 result = self._provide(args, kwargs)
             except Exception as exc:
-                print(self)
+                logging.debug(str(self))
                 raise exc
 
         if self.is_async_mode_disabled():


whysage:
@rmk135
Please review PR.
I'll not sure about it.

1. 
Here - https://github.com/ets-labs/python-dependency-injector/pull/510 - you write that after reraise exception all exceptions will be just general 'Exception'.
I test my code by adding addition try here
src/dependency_injector/providers.pxd
line 582
` 
try:
        return call(*args, **kwargs)
    except Exception as exception:
        exception.__traceback__.tb_frame.f_locals.clear()
        exception.__traceback__.tb_frame.f_locals.update(locals())
        raise exception`
   ` 

and receive type of Exception was TypeError (I used  this demo
https://github.com/ets-labs/python-dependency-injector/blob/master/examples/demo/with_di.py
without env var timeout - so error was - TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType')


2. Also, I'm not sure about printing providers, May be can log it somewhere.

Please advise.

Addition idea about problem was to add ENV var like DEPENDENCY_INJECTION_DEBUG_MODE
and print called providers only it is enabled.

I think this issue is very important - for me, it is real hell to find error in dozens providers by unclear error message.
Please help :-)

src/dependency_injector/providers.pyx
```diff
@@ -204,7 +204,11 @@ cdef class Provider(object):
         if self.__last_overriding is not None:
             result = self.__last_overriding(*args, **kwargs)
         else:
-            result = self._provide(args, kwargs)
+            try:
+                result = self._provide(args, kwargs)
+            except Exception as exc:
+                print(self)
```
rmk135, line 210: I'm not quite sure about using `print()` here. It will send some data to stdout, which might not always be the desired behavior. 

The idea, generally, looks interesting. Maybe we can use the ``logging`` module?

gen-xu: I would agree with @rmk135 to use the `logging` module with `logger.exception` with some extra useful messages as it allows users to have more control on the output through `logging.config`. Meanwhile, I believe the [LogRecord](https://docs.python.org/3/library/logging.html#logging.LogRecord) may not provide precise info given the cython context (caller line number, stacktrace etc.)

whysage: Thank you for reply!

@rmk135 @gen-xu 

Please look at updated pull request.
I hope logging.debug() is good idea because, by default, we have warning log level, and I hope no one use debug level in production) - so nothing will change for current users until logging level set to debug.
If this request will be accepted - I'll add article to documentation about this.

whysage:
Hi
@rmk135 @gen-xu 
Any updates?

whysage:
Hi!
@rmk135 @gen-xu
ping for updates)

PR #580, opened 2022-04-09
Fix a few issues in the introduction

illia-v:
Nice project, @rmk135!

I found a few issues when I read the introduction:
* "When the cohesion is high the coupling is low." – this is not always true, see [the article](https://enterprisecraftsmanship.com/posts/cohesion-coupling-difference/#_types_of_code_from_a_cohesion_and_coupling_perspective);
* some typing issues: passing a string when an integer is expected and missing return types when arguments are annotated.

### DIFF
--- a/docs/introduction/di_in_python.rst
+++ b/docs/introduction/di_in_python.rst
@@ -49,7 +49,8 @@ Coupling and cohesion are about how tough the components are tied.
 - **High cohesion**. High cohesion is like using the screws. Very easy to disassemble and
   assemble back or assemble a different way. It is an opposite to high coupling.
 
-High cohesion often correlates with loose coupling, and vice versa.
+Cohesion often correlates with coupling. Higher cohesion usually leads to lower coupling, and vice
+versa.
 
 Low coupling brings a flexibility. Your code becomes easier to change and test.
 
--- a/docs/introduction/di_in_python.rst
+++ b/docs/introduction/di_in_python.rst
@@ -49,7 +49,7 @@ Coupling and cohesion are about how tough the components are tied.
 - **High cohesion**. High cohesion is like using the screws. Very easy to disassemble and
   assemble back or assemble a different way. It is an opposite to high coupling.
 
-When the cohesion is high the coupling is low.
+High cohesion often correlates with loose coupling, and vice versa.
 
 Low coupling brings a flexibility. Your code becomes easier to change and test.
 
--- a/docs/introduction/di_in_python.rst
+++ b/docs/introduction/di_in_python.rst
@@ -66,14 +66,14 @@ Before:
 
    class ApiClient:
 
-       def __init__(self):
+       def __init__(self) -> None:
            self.api_key = os.getenv("API_KEY")  # <-- dependency
            self.timeout = int(os.getenv("TIMEOUT"))  # <-- dependency
 
 
    class Service:
 
-       def __init__(self):
+       def __init__(self) -> None:
            self.api_client = ApiClient()  # <-- dependency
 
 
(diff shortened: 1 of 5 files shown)


docs/introduction/di_in_python.rst
```diff
@@ -49,7 +49,7 @@ Coupling and cohesion are about how tough the components are tied.
 - **High cohesion**. High cohesion is like using the screws. Very easy to disassemble and
   assemble back or assemble a different way. It is an opposite to high coupling.
 
-When the cohesion is high the coupling is low.
+High cohesion often correlates with loose coupling, and vice versa.
```
rmk135, line 52: What do you think about something like that:

“Cohesion often correlates with coupling. Higher cohesion usually leads to lower coupling, and vice versa.”

rmk135:
Hi @illia-v ,

Thanks a lot for the PR and provided link.

As of the PR, could I please ask you to change the destination branch to `develop`?

Also, if I’m not wrong the same code sample is used on docs index page, on readme, and also there’s example code somewhere in the “/examples/“ folder. It would be nice to make the typing updates there as well. Please, let me know if it’s something you could help with.

Speaking about the thoughts on correlation of coupling and cohesion, I’m good to change the wording to the softer form, but generally I would say that it’s a separate discussion. For instance, when I think about God Object, I imagine a class with 30-50 methods calling each other in chaotic order and I cannot imagine if there’s room to call it highly cohesive & highly coupled structure — for me it looks just like an example of high coupling. 

Same about destructive cohesion, I totally agree about the term, reasoning, consequences and everything else, but I’m not sure if it’s any special type of correlation. I would say that it’s just an extreme level of cohesion, where extremely low level of coupling actually creates a problem.

PS: I’m happy to have a discussion about that, but probably would stop here for now :)

docs/introduction/di_in_python.rst
illia-v: LGTM! Changed in f8ec061

illia-v:
> As of the PR, could I please ask you to change the destination branch to `develop`?

Sure, done.

> Also, if I’m not wrong the same code sample is used on docs index page, on readme, and also there’s example code somewhere in the “/examples/“ folder. It would be nice to make the typing updates there as well. Please, let me know if it’s something you could help with.

Done.

---

> Speaking about the thoughts on correlation of coupling and cohesion, I’m good to change the wording to the softer form, but generally I would say that it’s a separate discussion. For instance, when I think about God Object, I imagine a class with 30-50 methods calling each other in chaotic order and I cannot imagine if there’s room to call it highly cohesive & highly coupled structure — for me it looks just like an example of high coupling.
> 
> Same about destructive cohesion, I totally agree about the term, reasoning, consequences and everything else, but I’m not sure if it’s any special type of correlation. I would say that it’s just an extreme level of cohesion, where extremely low level of coupling actually creates a problem.
> 
> PS: I’m happy to have a discussion about that, but probably would stop here for now :)

There was [a discussion](https://disq.us/p/2c4u5ht) in comments of the article about God object. It looks unfinished, unfortunately.

The article author defines  "the number of connections inside a code component (cohesion) and between code components (coupling)." If this definition is used, a God object does seem to have high cohesion, and destructive decoupling does seem to result in low cohesion.

BTW, there is [an article](https://blog.ttulka.com/how-cohesion-and-coupling-correlate) from his opponent, he provides some other examples of low cohesion + loose coupling and high cohesion + tight coupling.

Speaking for myself, I am still defining cohesion and coupling 🙂. They are pretty subjective.

rmk135 (approved):

rmk135:
@illia-v many thanks for the changes, merged. Also, thanks a lot for some education about cohesion and coupling, super helpful to look at things from various perspectives :) I'll definitely mention this PR when speak about coupling and cohesion next time.

PR #635, opened 2022-11-12
fixed bug in windows build with default charset

kirill-shershen:
I got some problems when i tried to build the package on windows 11 with cp1251 encoding by default.
`UnicodeDecodeError: 'charmap' codec can't decode byte 0x98 in position 8667: character maps to <undefined>`


rmk135 (approved):

rmk135:
@kirill-shershen thanks a lot for the contribution!

Issue #632, opened 2022-10-26
Python 3.11 support


gen-xu:
python 3.11 is released on 2022-10-24 this Monday. Current pip install with Python 3.11 fails and gives this failure
```python
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib.linux-x86_64-cpython-311
      creating build/lib.linux-x86_64-cpython-311/dependency_injector
      copying src/dependency_injector/__init__.py -> build/lib.linux-x86_64-cpython-311/dependency_injector
      copying src/dependency_injector/schema.py -> build/lib.linux-x86_64-cpython-311/dependency_injector
      copying src/dependency_injector/wiring.py -> build/lib.linux-x86_64-cpython-311/dependency_injector
      copying src/dependency_injector/resources.py -> build/lib.linux-x86_64-cpython-311/dependency_injector
      copying src/dependency_injector/errors.py -> build/lib.linux-x86_64-cpython-311/dependency_injector
      creating build/lib.linux-x86_64-cpython-311/dependency_injector/ext
      copying src/dependency_injector/ext/__init__.py -> build/lib.linux-x86_64-cpython-311/dependency_injector/ext
      copying src/dependency_injector/ext/aiohttp.py -> build/lib.linux-x86_64-cpython-311/dependency_injector/ext
      copying src/dependency_injector/ext/flask.py -> build/lib.linux-x86_64-cpython-311/dependency_injector/ext
      copying src/dependency_injector/containers.pxd -> build/lib.linux-x86_64-cpython-311/dependency_injector
      copying src/dependency_injector/providers.pxd -> build/lib.linux-x86_64-cpython-311/dependency_injector
      copying src/dependency_injector/providers.pyi -> build/lib.linux-x86_64-cpython-311/dependency_injector
      copying src/dependency_injector/__init__.pyi -> build/lib.linux-x86_64-cpython-311/dependency_injector
      copying src/dependency_injector/containers.pyi -> build/lib.linux-x86_64-cpython-311/dependency_injector
      copying src/dependency_injector/py.typed -> build/lib.linux-x86_64-cpython-311/dependency_injector
      running build_ext
      building 'dependency_injector.containers' extension
      creating build/temp.linux-x86_64-cpython-311
      creating build/temp.linux-x86_64-cpython-311/src
      creating build/temp.linux-x86_64-cpython-311/src/dependency_injector
[...]

PR #646, opened 2022-12-18
feat: re cythonize to support python 3.11

gen-xu:
This PR fixes https://github.com/ets-labs/python-dependency-injector/issues/632


rmk135:
Hey @gen-xu ! Thanks a lot for the contribution! I'm merging into the existing `python3.11` branch and will make more changes to the build system. The problem right now is that Python 3.5 - 3.6 builds are not available in Ubuntu 22.04. I'll try to split the matrix and migrate to 22.04 only newer versions.

PR #673, opened 2023-02-28
Fix broken wiring of sync inject-decorated methods

### LINKED ISSUE #672: Broken wiring of sync inject-decorated methods
Hi,

following use case seems broken in versions >= 4.39.0:
```python
from dependency_injector.wiring import inject, Provide


class Service:
    @staticmethod
    def run():
        print("hello world!")


    service = providers.Factory(Service)


class MyClass:
    @inject
    def __init__(self, service=Provide[Container.service]):
        service.run()


container = Container()
container.wire(modules=[MyClass])
MyClass()
```
Output in version 4.38.0:
```
hello world!
```
Output in versions >4.38.0
```
Traceback (most recent call last):
  File "C:\Users\myself\file.py", line 23, in <module>
    MyClass()
  File "src/dependency_injector/_cwiring.pyx", line 28, in dependency_injector._cwiring._get_sync_patched._patched
  File "C:\Users\myself\file.py", line 18, in __init__
    service.run()
AttributeError: 'Provide' object has no attribute 'run'
```

What seem to happen is that since the `_get_sync_patched` decorator has been moved to the cython part of the package, the signature of the `@inject`-decorated func/methods are now cython (`type(MyClass.__init__)` resolves to `cython_function_or_method`), therefore `inspect.isfunction()`/`inspect.ismethod()` resolve to False, and so those functions are never patched during container wiring.

Might be related to [#589](https://github.com/ets-labs/python-dependency-injector/issues/589).

Finally, I have successfully tested [here](https://github.com/ets-labs/python-dependency-injector/pull/673) using the same patch that was applied in 4.39.1 for `_get_async_patched` coroutine that wasn't properly signed from cython coroutine, so creating the `_get_sync_patched` in wiring.py that calls _cwiring's `_sync_inject`, which effectively conserves behavior when running wiring.

martlaf:
This PR fixes #672


javiermrzfever:
Any news on this PR??

ZipFile (approved):

PR #683, opened 2023-03-21
Add Starlette lifespan handler implementation

ZipFile:
Starlette recently added cool new feature called [lifespan](https://www.starlette.io/lifespan/) in [v0.26.0](https://github.com/encode/starlette/releases/tag/0.26.0) and FastAPI [followed the suit](https://fastapi.tiangolo.com/advanced/events/#lifespan) in [v0.93.0](https://github.com/tiangolo/fastapi/releases/tag/0.93.0). This makes it possible to get rid of good chunk of initialization spaghetti in starlette-based apps (making container a single entry point in the application), + reduces amount of the pain in the ass scheduling background tasks in the event loop of the ASGI app.

<details>
  <summary>Example</summary>

  `myapp/container.py`
  ```python
  from dependency_injector.ext.starlette import Lifespan
  from dependency_injector.providers import Factory, Resource, Self, Singleton
  from fastapi import FastAPI
  
  
  def init():
    print("Inittializing resources")
    yield
    print("Cleaning up resources")
  
  
  class Container(DeclarativeContainer):
      __self__ = Self()
      lifespan = Singleton(Lifespan, __self__)
      app = Factory(FastAPI, lifespan=lifespan)
      init = Resource(init)
  ```
  `myapp/asgi.py`:
  ```python
  from .container import Container
  
  container = Container()
  app = container.app()
  ```
  `run.sh`:
  ```sh
  uvicorn myapp.asgi:app
  ```
</details>

### DIFF
--- a/src/dependency_injector/ext/starlette.py
+++ b/src/dependency_injector/ext/starlette.py
@@ -1,0 +1,53 @@
+import sys
+from abc import ABCMeta, abstractmethod
+from typing import Any, Callable, Coroutine, Optional
+
+if sys.version_info >= (3, 11):  # pragma: no cover
+    from typing import Self
+else:  # pragma: no cover
+    from typing_extensions import Self
+
+from dependency_injector.containers import Container
+
+
+class Lifespan:
+    """A starlette lifespan handler performing container resource initialization and shutdown.
+
+    See https://www.starlette.io/lifespan/ for details.
+
+    Usage:
+
+    .. code-block:: python
+
+        from dependency_injector.containers import DeclarativeContainer
+        from dependency_injector.ext.starlette import Lifespan
+        from dependency_injector.providers import Factory, Self, Singleton
+        from starlette.applications import Starlette
+
+        class Container(DeclarativeContainer):
+            __self__ = Self()
+            lifespan = Singleton(Lifespan, __self__)
+            app = Factory(Starlette, lifespan=lifespan)
+
+    :param container: container instance
+    """
+
+    container: Container
+
+    def __init__(self, container: Container) -> None:
+        self.container = container
+
+    def __call__(self, app: Any) -> Self:
+        return self
+
+    async def __aenter__(self) -> None:
+        result = self.container.init_resources()
+
+        if result is not None:
+            await result
+
+    async def __aexit__(self, *exc_info: Any) -> None:
+        result = self.container.shutdown_resources()
+
+        if result is not None:
+            await result
--- a/examples/miniapps/starlette-lifespan/README.rst
+++ b/examples/miniapps/starlette-lifespan/README.rst
@@ -1,0 +1,39 @@
+Integration With Starlette-based Frameworks
+===========================================
+
+This is a `Starlette <https://www.starlette.io/>`_ +
+`Dependency Injector <https://python-dependency-injector.ets-labs.org/>`_ example application
+utilizing `lifespan API <https://www.starlette.io/lifespan/>`_.
+
+.. note::
+
+    Pretty much `any framework built on top of Starlette <https://www.starlette.io/third-party-packages/#frameworks>`_
+    supports this feature (`FastAPI <https://fastapi.tiangolo.com/advanced/events/#lifespan>`_,
+    `Xpresso <https://xpresso-api.dev/latest/tutorial/lifespan/>`_, etc...).
+
+Run
+---
+
+Create virtual environment:
+
+.. code-block:: bash
+
+    python -m venv env
+    . env/bin/activate
+
+Install requirements:
+
+.. code-block:: bash
+
+    pip install -r requirements.txt
+
+To run the application do:
+
+.. code-block:: bash
+
+    python example.py
+    # or (logging won't be configured):
+    uvicorn --factory example:container.app
+
+After that visit http://127.0.0.1:8000/ in your browser or use CLI command (``curl``, ``httpie``,
+etc).
--- a/examples/miniapps/starlette-lifespan/example.py
+++ b/examples/miniapps/starlette-lifespan/example.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python
+
+from logging import basicConfig, getLogger
+
+from dependency_injector.containers import DeclarativeContainer
+from dependency_injector.ext.starlette import Lifespan
+from dependency_injector.providers import Factory, Resource, Self, Singleton
+from starlette.applications import Starlette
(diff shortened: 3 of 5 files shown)


Trinkes:
Will this the scope the life span of a container to a request?
If I'm understanding this correctly, it will make it very easy to have a per request database session, right?
Should we update the [fastapi](https://python-dependency-injector.ets-labs.org/examples/fastapi-sqlalchemy.html?highlight=fastapi) example to include this approach?

ZipFile:
Looks like project is still alive. Rebased with latest `master`.

src/dependency_injector/ext/starlette.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __call__(self, app: Any) -> Self:
+        return self
+
+    async def __aenter__(self) -> None:
+        result = self.container.init_resources()
+
+        if result is not None:
+            await result
+
```
nightblure, line 48: Hi!
Thanks for the nice change. Perhaps it would be appropriate to add an explicit `return None` at the end of `__aenter__`

src/dependency_injector/ext/starlette.py
```diff
@@ -0,0 +1,53 @@
...
+    container: Container
+
+    def __init__(self, container: Container) -> None:
+        self.container = container
+
+    def __call__(self, app: Any) -> Self:
+        return self
+
+    async def __aenter__(self) -> None:
```
nightblure, line 43: Here you need to change the return type probably to `Optional[Awaitable]`

ZipFile: 1. This method will be inferred as `def __aenter__(self) -> Awaitable[None]:`
2. `-> Optional[Awaitable]` means function returns either `None` or something that you can `await` (e.g. signature of the `init_resources`). Given the pt.1, annotating it with your suggestion will result in `Awaitable[Optional[Awaitable]]`, which does not conforms to [StatelessLifespan](https://github.com/encode/starlette/blob/c2e3a39b09a613553ee03586589ed9cd0fbf07f3/starlette/types.py#L18) protocol.

src/dependency_injector/ext/starlette.py
ZipFile: You generally want explicit `return None` at the end if you `return`ed something earlier in the function. There are no returns at all here, so this is extra.

VladyslavHl:
Great! Integrating DI Resources with FastAPI lifespan is something me and colleagues will investigate soon, good to see the potential of universal solution.

IMO it would be benefitial to make this feature more flexible, allowing user to chose what resources to initialize in lifespan as you may want to init some resources at other moment 

From the top of my head Lifespan can take individual resources to be initialized/shutdown on lifespan events, so the usage could be:

```python
class Container(DeclarativeContainer):
    init = Resource(init)
    other_resource = Resource(init_other_resource)   # to be intialized not on startup event
    lifespan = Singleton(Lifespan, init, ... )
    app = Factory(FastAPI, lifespan=lifespan)
    ...
```
also preserving the option to pass `__self__` to init all resources of container as it is implemented now.

This leads to question how to not initialize these resources when calling `container.init_resources()`, maybe the its sufficient to just init them explicily by calling `container.other_resource.init()` when needed.

maybe I'm complicating, let me know wdyt

ZipFile:
@VladyslavHl I guess your problem is much more general and your proposal will only solve it partially if at all.

From my experience using DI, it lacks notion of scopes for resources. Like "app", "request", "task", etc... I've solved the problem by doing scopes explicitly, i.e. injecting context manager that will initialize resources (e.g. db sessions) and storing them into contextvars. Not exactly great solution, but works when you have mix of different "apps" in your project (fastapi, typer, celery, alembic, etc...) but still need common setup code to run (configuring logging, apms, service discovery, etc...).

Introducing scopes would probably require significant re-engeneering of how DI itself works, so for the time being my advice for you would be to implement your own `Lifespan` the way you proposed. My implementation is intentionally dumb.

Also, take a look at [ContextLocalSingleton](https://github.com/ets-labs/python-dependency-injector/blob/abf2a2577c6e6d1cfd5b9a6ee48d26501fc18c7a/src/dependency_injector/providers.pyx#L3231-L3248). This is not a 100% replacement for resources, but might help in certain situations.

rmk135 (approved):

Totorokrut:
Any chance to get it work with DynamicContainer?

I took exampel from the docs and changed container type to the dynamic.
```python
from contextlib import asynccontextmanager
from dependency_injector.ext.starlette import Lifespan
from fastapi import FastAPI, Request, Depends, APIRouter


class Connection: ...


@asynccontextmanager
async def init_database():
    print("opening database connection")
    yield Connection()
    print("closing database connection")


router = APIRouter()


@router.get("/")
@inject
async def index(request: Request, db: Connection = Depends(Provide["db"])):
    # use the database connection here
    return "OK!"


class Container2(containers.DynamicContainer):
    __self__ = providers.Self()
    db = providers.Resource(init_database)
    lifespan = providers.Singleton(Lifespan, __self__)
    app = providers.Singleton(FastAPI, lifespan=lifespan)
    _include_router = providers.Resource(
        app.provided.include_router.call(),
        router,
    )


if __name__ == "__main__":
    import uvicorn

    container = Container2()
    app = container.app()
    uvicorn.run(app, host="localhost", port=8000)
```
And it fails now with:
```
INFO:     Started server process [878575]
INFO:     Waiting for application startup.
ERROR:    Traceback (most recent call last):
  File "/home/lglassner/Work/listener/.venv/lib/python3.12/site-packages/starlette/routing.py", line 692, in lifespan
    async with self.lifespan_context(app) as maybe_state:
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lglassner/Work/listener/.venv/lib/python3.12/site-packages/dependency_injector/ext/starlette.py", line 52, in __aenter__
    result = self.container.init_resources(self.resource_type)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'init_resources'

ERROR:    Application startup failed. Exiting.
```
The issue is there is no \_\_self__ for dynamic container as I understood. As in moment of creaeting Lifespan in \_\_init__ there is just container=None.

Maybe we can somehow link to an existing container at run time?
Something like that:
```python
if __name__ == "__main__":
    import uvicorn

    container = Container2()
    container.__self__ = container
    app = container.app()
    uvicorn.run(app, host="localhost", port=8000)
```

PR #717, opened 2023-06-27
Fix docker image version in asyncio daemon tutorial

jonatasleon:
In asyncio daemon tutorial, at the part before define the `Dockerfile` content, it's said it will use `python:3.9-buster`, but in `Dockerfile`, it uses `python:3.10-buster` in `FROM` tag.

This PR proposes set both as `python:3.10-buster` image version.


ZipFile:
Hey, thanks for PR. Documentation was updated as part of upgrading deps in #838. So I'm closing this PR since this change is no longer relevant.

Issue #731, opened 2023-08-03
type safety for provided?


philipbjorge:
When referencing a singleton via `foo.provided.bar`, we lose type safety because `provided` always resolves to the `ProvidedInstance` rather than the specific type of the dependency.

```
    settings = providers.Singleton(Settings)
    
    identity_container = providers.Container(
        IdentityContainer,
        base_url=settings.provided.IDENTITY_SERVICE_BASE_URL,
    )
```

**Expected**
`settings.provided` resolves to `Settings`

**Actual**
`settings.provided` resolves to `ProvidedInstance`

This seems like an easy change to make -- Is there any reason this wouldn't work if I opened a PR?

toinbis:
@philipbjorge  - installed your fix from git, works like magic! @rmk135 is there any chance we could get it merge?

philipbjorge:
@toinbis -- Another type update :)
https://github.com/ets-labs/python-dependency-injector/pull/744

ZipFile:
Available in [v4.46.0](https://github.com/ets-labs/python-dependency-injector/releases/tag/4.46.0).

PR #733, opened 2023-08-03
fix: type propogation through provided

philipbjorge:
Addresses https://github.com/ets-labs/python-dependency-injector/issues/731


jonathangreen:
@rmk135 Any chance that this type fix could be merged as well?

rmk135:
Thank you for the contribution @philipbjorge .

@ZipFile can you please review this PR when you have a moment?

tests/typing/callable.py
```diff
@@ -34,10 +34,7 @@ def create(cls) -> Animal:
 
 # Test 5: to check the provided instance interface
 provider5 = providers.Callable(Animal)
-provided5: providers.ProvidedInstance = provider5.provided
-attr_getter5: providers.AttributeGetter = provider5.provided.attr
-item_getter5: providers.ItemGetter = provider5.provided["item"]
-method_caller: providers.MethodCaller = provider5.provided.method.call(123, arg=324)
+provided5: Animal = provider5.provided
```
rmk135, line 37: This test is not exactly correct. The attribute ``provider5.provided`` returns a provider, not an object of ``Animal``. I can't recall why it was important, but the current change will introduce a functional change in the typing approach.

To make a correct typing test with the ``.provided`` attribute we would need to do something like that:

```python
provided5: Animal = provider5.provided()
```

At least, that would be what the library actually does.

philipbjorge:
I don't have capacity to address any feedback here and no longer work at the company using this library -- But I've reached out to my previous teammates to see if they can carry the torch :)

Thanks for reviewing! Excited to see this project waking back up!

tests/typing/callable.py
GonzaGonza: @rmk135 updated the tests with `.provided()`

cc @philipbjorge thanks for flagging

src/dependency_injector/providers.pyi
```diff
@@ -69,7 +69,7 @@ class Provider(Generic[T]):
     @property
     def provider(self) -> Provider: ...
     @property
-    def provided(self) -> ProvidedInstance: ...
+    def provided(self) -> T: ...
```
ZipFile, line 72:
```suggestion
    def provided(self) -> ProvidedInstance[T]: ...
```

tests/typing/callable.py
```diff
@@ -34,10 +34,7 @@ def create(cls) -> Animal:
 
 # Test 5: to check the provided instance interface
 provider5 = providers.Callable(Animal)
-provided5: providers.ProvidedInstance = provider5.provided
-attr_getter5: providers.AttributeGetter = provider5.provided.attr
-item_getter5: providers.ItemGetter = provider5.provided["item"]
-method_caller: providers.MethodCaller = provider5.provided.method.call(123, arg=324)
+provided5: Animal = provider5.provided()
```
ZipFile, line 37: If you run `mypy` against this code you'll get an error `tests/typing/callable.py:37: error: "Animal" not callable  [operator]`.

tests/typing/callable.py
```diff
@@ -34,10 +34,7 @@ def create(cls) -> Animal:
 
 # Test 5: to check the provided instance interface
 provider5 = providers.Callable(Animal)
-provided5: providers.ProvidedInstance = provider5.provided
-attr_getter5: providers.AttributeGetter = provider5.provided.attr
-item_getter5: providers.ItemGetter = provider5.provided["item"]
-method_caller: providers.MethodCaller = provider5.provided.method.call(123, arg=324)
```
ZipFile, line 40: I think we should not remove this, since `provided` attribute is in fact `ProvidedInstance`:
https://github.com/ets-labs/python-dependency-injector/blob/46646b1acfd833eeb55075651af9a13fcb4762ef/src/dependency_injector/providers.pyx#L431-L434

ZipFile (changes requested):

tests/typing/callable.py
GonzaGonza: @ZipFile I looked into this and I think you are right, I reintroduced the removed tests.

ZipFile (approved):
LGTM

PR #744, opened 2023-09-08
fix: propogate type through Provider

philipbjorge:
**Before**
CleanShot 2023-09-08 at 14 46 06

**After**
CleanShot 2023-09-08 at 14 45 45


jonathangreen:
@rmk135 this seems like a big improvement for type safety, any change of this getting merged in?

rmk135:
@philipbjorge thank you for the contribution. This change looks good to me. I'm not sure why there was no generic propagation. Looks like a good bugfix.

There also should be some tests for that. Maybe we will need to fix them. Not sure why CI is not running checks for this PR.

rmk135 (approved):

rmk135:
@ZipFile can you please review this PR as well?

philipbjorge:
I don't have capacity to address any feedback here and no longer work at the company using this library -- But I've reached out to my previous teammates to see if they can carry the torch :)

Thanks for reviewing! Excited to see this project waking back up!

ZipFile (commented):
Looks good to me either, but it would be nice to have some tests too.

PR #756, opened 2023-10-26
Add additional version of Resources

cpvandehey:
It would be really useful to have a `Resource` type that is capable of having a sync `init` function with an async `shutdown`.

There are many scenarios where this is useful

Ex:

1. Create an [aiohttp client](https://docs.aiohttp.org/en/stable/client_quickstart.html#make-a-request).
`session = aiohttp.ClientSession()`

2. Before closing the app, I want to await any hanging connections
`await session.close()`

### DIFF
--- a/src/dependency_injector/resources.py
+++ b/src/dependency_injector/resources.py
@@ -25,3 +25,13 @@ class AsyncResource(Generic[T], metaclass=abc.ABCMeta):
 
     async def shutdown(self, resource: Optional[T]) -> None:
         ...
+
+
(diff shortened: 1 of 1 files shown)


ZipFile:
I do not really think we need specific variations of `AsyncResource` (sync init + async shutdown, async init + sync shutdown), since the way how `async` stuff works in Python. `async def` function that never awaits have performance of the regular generator (see [docs](https://docs.python.org/3/reference/datamodel.html#coroutine-objects) on coroutines), since it never switches to the event loop.

Issue #681, opened 2023-03-20
Ressource not closed in case of Exception in the function


BaptisteSaves:
Hello,

We have some function that uses `Closing[Provide["session"]]` to inject the database session to the method.
It works well, except if we have any exception raised in the function the closing method will not be called and the resource will stay open

This is really easy to reproduce with a minimal case on fastAPI, the shudown (`db.close()`) is never called in this case.

This is a huge problem as in the next call, the injector will re inject the resource (the session) which is already in a fail state.

I did not find anything on the documentation regarding error management, did we miss something ?

Thanks for your help. 

Container
```
class Container(containers.DeclarativeContainer):
    wiring_config = containers.WiringConfiguration(packages=[XX])

    session = providers.Resource(get_db)
```

Resource DB:
```
def get_db() -> Iterator[Session]:
    db: Session = SessionMaker()
    try:
        yield db
    finally:
        db.close()
```

Router:
```
@router.get(
    "/hello-world"
)
@inject
async def hello_world(
    session: Session = Depends(Closing[Provide["session"]]),
):
    raise Exception()
```

ZipFile:
Fixed in v4.48.0.

PR #769, opened 2023-11-28
Close the resources in case there was an exception

ivanovart:
I have the same problem as in https://github.com/ets-labs/python-dependency-injector/issues/681: the resources are not closed in case functions are failing, and I have my exceptions handler on a higher level then functions I inject.
So I suggest the change to close the dependencies in any case.

### DIFF
--- a/src/dependency_injector/_cwiring.c
+++ b/src/dependency_injector/_cwiring.c
@@ -1742,6 +1742,16 @@ struct __pyx_obj_19dependency_injector_8_cwiring___pyx_scope_struct_1__async_inj
   PyObject *__pyx_v_to_close_await;
   PyObject *__pyx_v_to_inject;
   PyObject *__pyx_v_to_inject_await;
+  int __pyx_t_0;
+  int __pyx_t_1;
+  char const *__pyx_t_2;
+  PyObject *__pyx_t_3;
+  PyObject *__pyx_t_4;
+  PyObject *__pyx_t_5;
+  PyObject *__pyx_t_6;
+  PyObject *__pyx_t_7;
+  PyObject *__pyx_t_8;
+  PyObject *__pyx_t_9;
 };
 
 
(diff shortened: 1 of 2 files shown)


src/dependency_injector/_cwiring.pyx
```diff
@@ -62,25 +62,24 @@ async def _async_inject(object fn, tuple args, dict kwargs, dict injections, dic
...
-                to_close_await.append(shutdown)
-
-        await asyncio.gather(*to_close_await)
+            await asyncio.gather(*to_close_await)
 
-    return result
 
 
-cdef bint _isawaitable(object instance):
+cdef bint _is_awaitable(object instance):
```
ZipFile, line 82: This change looks unrelated to the main topic of PR. Please revert to avoid diff noise.

src/dependency_injector/_cwiring.pyx
```diff
@@ -25,17 +25,17 @@ def _get_sync_patched(fn, patched: PatchedCallable):
             if arg_key not in kwargs or isinstance(kwargs[arg_key], _Marker):
                 to_inject[arg_key] = provider()
 
-        result = fn(*args, **to_inject)
+        try:
+            return fn(*args, **to_inject)
+        finally:
```
ZipFile, line 30: We need a test case to verify that this code block now is always executed.

src/dependency_injector/_cwiring.pyx
```diff
@@ -62,25 +62,24 @@ async def _async_inject(object fn, tuple args, dict kwargs, dict injections, dic
         async_to_inject = await asyncio.gather(*(provide for _, provide in to_inject_await))
         for provide, (injection, _) in zip(async_to_inject, to_inject_await):
             to_inject[injection] = provide
+    try:
+        return await fn(*args, **to_inject)
+    finally:
```
ZipFile, line 67: See comment above.

src/dependency_injector/_cwiring.c
ZipFile, line 1: We no longer store generated C files in source tree. Rebasing with latest `master` should no longer produce merge conflicts.

ZipFile (changes requested):
Thanks for contribution! Looks promising, but it needs additional work to do.

Issue #787, opened 2024-02-22
Error using toml, dependency -injector with fastapi and python3.12


1988gadocansey:
Note: This error originates from the build backend, and is likely not a problem with poetry but with dependency-injector (4.41.0) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "dependency-injector (==4.41.0)"'.

using
 toml, dependency -injector 
 fastapi and python3.12

Issue #771, opened 2023-12-18
Error when installing by poetry in python 3.12


DaggerJackfast:
I have the error when install the package with poetry in python 3.12.1
`Note: This error originates from the build backend, and is likely not a problem with poetry but with dependency-injector (4.41.0) not supporting PEP 517 builds. You can verify this by running 'pip wheel --use-pep517 "dependency-injector (==4.41.0)"'.`

changja88:
I have same problem.
Python 3.12 is super cool. I hope I can use dependency-injector with python 3.12 as soon as possible.
Thanks!

yaziamoe:
I like to have it enabled for Python 3.12 too. My project is stuck using this amazing framework to python 3.11. 

Thanks in advance!

sh-at-cs:
The PR for adding Python 3.12 support is https://github.com/ets-labs/python-dependency-injector/pull/765

rmk135:
The release with Python 3.12 support is on PyPI: https://pypi.org/project/dependency-injector/4.42.0b1/

```
pip install dependency-injector==4.42.0b1
```

Please report any issues you notice by opening an issue here: https://github.com/ets-labs/python-dependency-injector/issues

I'll publish the plan for future steps a bit later. There is a lot of work to do.

ZipFile:
Latest release is now fully [PEP-517](https://peps.python.org/pep-0517/)-compliant (see #829) and should be installable by modern versions of pip or alternative package mangers like Poetry. I've verified that it installs fine on Python 3.12 using Poetry. Closing.

PR #789, opened 2024-03-02
regenerate .c files

hp-peti:
using Cython 0.29.37

this enables compilation with cpython 3.12 header

nothing else was changed


hp-peti:
- pyproject.toml-based installations work
- usage with betterproto and grpclib was successful 
- not tested with python <3.12 though

hp-peti:
might fix #787

hp-peti:
might fix #771 as well

ZipFile:
We no longer store generated C files in the source tree, so this PR is no longer relevant. Latest release should be compatible with recent Python versions either way, so clothing.

Issue #812, opened 2024-08-13
Roadmap 2024


rmk135:
### 1. Add Python 3.12 support (beta)

Status: done

Description:

- Dependency Injector is based on Cython version 0.29. Unit test coverage has never worked on Python 3.12 with the most recent version of Cython 0.29. There were compilation errors. The culprit is C API incompatibilities in the Cython code. Since there were many C API changes in Python 3.12 and the Cython project moved its effort towards their 3.* branch, the current Cython code might be deficient. I decided to publish it as a beta version to prevent automatic updates for people who haven’t pinned their versions.
- There have been no issues regarding the beta release. It will be promoted to a full release in a few weeks. 

### 2. Add Python 3.12 support (release)

Status: in progress

Description:
- See the previous item 

### 3. Upgrade to Cython 3

Status: to do 

Description:
- The branch of Cython that Dependency Injector is on (0.29) won't get any new releases
- Currently, the Dependency Injector tests are crashing with segmentation fault errors in hundreds of tests
- I anticipate a lot of work here, something has changed significantly in Cython 3 and/or Python 3.12+ C API
- Updating to Cython 3 is a pre-requisite to adding Python 3.13 support 

### 4. Add Python 3.13

Status: to do

Description:

- The new version of Python is coming this fall

### 5. Update security vulnerabilities

Status: to do

Description:

- There are quite a few dependencies on other frameworks in the tutorials and other examples
- All those dependencies will be updated to the most recent versions

### 6. Operations and ongoing support

Status: ongoing

Description:

- This includes working with opened issues, pull requests, and bugs

### Sponsorship

I’ve decided to give a try to the GitHub Sponsors program. Delivering this roadmap is going to require hundreds of hours of work. If you would like to support me in this journey, please contribute here:

https://github.com/sponsors/rmk135

### Questions

If you have any questions or suggestions please open a new issue. This issue is going to be locked for leaving comments.

rmk135:
"2. Add Python 3.12 support (release)" has been done.

"3. Upgrade to Cython 3" is in progress.

rmk135:
"3. Upgrade to Cython 3" has been done by @ZipFile .

"4. Add Python 3.13" has been done and released to PyPI.

ZipFile:
I think we've done pretty much everything from this list. I have few ideas for 2026 roadmap, I'll get in touch with you later this month.

PR #813, opened 2024-08-13
Migrate to Cython3

ZipFile:
As title says. Major incompatibility with v0.29 was new mandatory dunder mangling, so solution was simply to rename them. Otherwise migration was pretty much straightforward.

Notes:
* Cython 3 no longer recommends commiting C files. I already have some drafts, will submit PR after this one is merged.
* So far it builds and runs on CPython 3.7-3.12 (with coverage).
* On PyPy it won't build with `DEPENDENCY_INJECTOR_DEBUG_MODE` (probably Cython issue), without debug mode looks like it builds and runs fine (no coverage ofc), but there are few failing tests, need investigate more if this is related to version of PyPy itself (I was testing with 7.3.16).
* Not exactly sure about segfaults mentioned in #812, but majority of tests were failing because Cython 3 started mangling attributes that were previously unmangled.


rmk135:
@ZipFile wow, thanks for the PR!

ZipFile:
Looks like tox was ignoring `DEPENDENCY_INJECTOR_DEBUG_MODE` env var set from GHA pipeline, coverage supposed to be around 94% with `*.pyx` included. Pushed fix + extra logging.

oleksandr-kuzmenko:
@ZipFile thanks for the PR, It looks cool 👏 

Hi @rmk135, would you mind taking a look at this PR when you have a chance? 🙏

Is there anything else needed to move it forward? I can try to help if needed

Issue #734, opened 2023-08-03
Unable to initialize instance that are picklable in the container


arunbalasubramani:
Getting the following error while initializing the container that contains instances that cannot be pickled. Is there a requirement that all the instances created within the container should be picklable?

Version used: 4.41.0

```
class DependencyContainer(containers.DeclarativeContainer):
    metric_flow_client = providers.Singleton(
        MetricFlowClient,
        sql_client=SnowflakeSqlClient.from_config(mf_snowflake_config)
    )

if __name__ == "__main__":
    container = DependencyContainer().   <---- Fails here
    container.init_resources()
    container.wire(modules=[__name__])
```

**Error log:**

`Traceback (most recent call last):
  File "/Users/arunkumar/Projects/unified-metrics-platform/consumption_grpc/grpc_server.py", line 79, in <module>
    container = DependencyContainer()
  File "src/dependency_injector/containers.pyx", line 730, in dependency_injector.containers.DeclarativeContainer.__new__
  File "src/dependency_injector/providers.pyx", line 4913, in dependency_injector.providers.deepcopy
  File "src/dependency_injector/providers.pyx", line 4920, in dependency_injector.providers.deepcopy
  File "/Users/arunkumar/.pyenv/versions/3.9.10/lib/python3.9/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/Users/arunkumar/.pyenv/versions/3.9.10/lib/python3.9/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/Users/arunkumar/.pyenv/versions/3.9.10/lib/python3.9/copy.py", line 153, in deepcopy
    y = copier(memo)
  File "src/dependency_injector/providers.pyx", line 2835, in dependency_injector.providers.BaseSingleton.__deepcopy__
  File "src/dependency_injector/providers.pyx", line 4920, in dependency_injector.providers.deepcopy
  File "/Users/arunkumar/.pyenv/versions/3.9.10/lib/python3.9/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/Users/arunkumar/.pyenv/versions/3.9.10/lib/python3.9/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/Users/arunkumar/.pyenv/versions/3.9.10/lib/python3.9/copy.py", line 172, in deepcopy
[...]

Issue #712, opened 2023-05-26
Django fails to start and throws TypeError


tkidwai-blink:
We recently refreshed dependencies of a django based application that uses dependency-injector v4.41.0. Whenever we try to go beyond django 3.1.14 (i.e. >= 3.2.0) we get the following error on startup of the application:

```
Exception in thread django-main-thread:
Traceback (most recent call last):
  File "/Users/t.k/.pyenv/versions/3.9.0/lib/python3.9/threading.py", line 950, in _bootstrap_inner
    self.run()
  File "/Users/t.k/.pyenv/versions/3.9.0/lib/python3.9/threading.py", line 888, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/t.k/Library/Caches/pypoetry/virtualenvs/prs-o9yqrpDn-py3.9/lib/python3.9/site-packages/django/utils/autoreload.py", line 64, in wrapper
    fn(*args, **kwargs)
  File "/Users/t.k/Library/Caches/pypoetry/virtualenvs/prs-o9yqrpDn-py3.9/lib/python3.9/site-packages/django/core/management/commands/runserver.py", line 110, in inner_run
    autoreload.raise_last_exception()
  File "/Users/t.k/Library/Caches/pypoetry/virtualenvs/prs-o9yqrpDn-py3.9/lib/python3.9/site-packages/django/utils/autoreload.py", line 87, in raise_last_exception
    raise _exception[1]
  File "/Users/t.k/Library/Caches/pypoetry/virtualenvs/prs-o9yqrpDn-py3.9/lib/python3.9/site-packages/django/core/management/__init__.py", line 375, in execute
    autoreload.check_errors(django.setup)()
  File "/Users/t.k/Library/Caches/pypoetry/virtualenvs/prs-o9yqrpDn-py3.9/lib/python3.9/site-packages/django/utils/autoreload.py", line 64, in wrapper
    fn(*args, **kwargs)
  File "/Users/t.k/Library/Caches/pypoetry/virtualenvs/prs-o9yqrpDn-py3.9/lib/python3.9/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/Users/t.k/Library/Caches/pypoetry/virtualenvs/prs-o9yqrpDn-py3.9/lib/python3.9/site-packages/django/apps/registry.py", line 122, in populate
    app_config.ready()
[...]

PR #839, opened 2024-12-14
Improve debugability of deepcopy errors

ZipFile:
This PR indirectly addresses issues mentioned in #712 and #734 by adding more specialized versions of `deepcopy` for providers that accept args and/or kwargs (Factory, List, etc...).

This is not a solution for the issues mentioned above, but it will help users to figure out what caused the error, since `deepcopy` stack traces are not informative at all.


rmk135 (approved):
Thank you, @ZipFile ! That's a good improvement indeed!

PR #845, opened 2024-12-18
:art: Raise exception instead of hiding it in finally

gortibaldik:
We use the dependency-injector a lot in our private project. Currently, when the configuration of the ThreadLocalSingleton is invalid (i.e. a constructor argument is forgotten), the library prints only this exception:

```
UnboundLocalError: local variable 'instance' referenced before assignment
```

This is really frustrating, because it provides no leads to the actual error. At the company, for last 7 months, we use a fork of this library, https://github.com/gortibaldik/dependency-injector-fg with a separate PyPI repo. However as we see that this project is again maintained, we would like to merge this small PR, which solves approx 50% of all of our problems with the library (the other 50% are issues with wiring which stem from our misunderstandings... :grinning: ).

### DIFF
--- a/src/dependency_injector/providers.pyx
+++ b/src/dependency_injector/providers.pyx
@@ -3221,8 +3221,8 @@ cdef class ThreadLocalSingleton(BaseSingleton):
                 return future_result
 
             self._storage.instance = instance
-        finally:
-            return instance
+        
+        return instance
 
     def _async_init_instance(self, future_result, result):
         try:
--- a/tests/unit/providers/singleton/test_thread_local_singleton_py3.py
+++ b/tests/unit/providers/singleton/test_thread_local_singleton_py3.py
@@ -1,0 +1,20 @@
+import pytest
+
+from dependency_injector.containers import Container
+from dependency_injector.providers import ThreadLocalSingleton
+
+
+class FailingClass:
+    def __init__(self):
+        raise ValueError("FAILING CLASS")
+
+
+class TestContainer(Container):
+    failing_class = ThreadLocalSingleton(FailingClass)
+
+
+def test_on_failure_value_error_is_raised():
+    container = TestContainer()
+
+    with pytest.raises(ValueError, match="FAILING CLASS"):
+        container.failing_class()


gortibaldik:
@rmk135, do you have time to look at this ?

On the other hand, we have another small issue. In your pipelines you do not upload wheels to the PyPI, which makes all the builds of our private library too long. In the repo discussed above (https://github.com/gortibaldik/dependency-injector-fg) we push wheels to the PyPI. What do you think about that ?

tests/unit/providers/singleton/test_thread_local_singleton_py2_py3.py
ZipFile, line 1: `py2` is extra here, we no longer support Python 2

ZipFile: Also run `isort` and `black` against this file. Code style.

ZipFile (commented):
Hey! Looks interesting, I'll check this more in depth later this week.

In the mean time, could you provide a real-world (or something that looks close to it) example of those 50% cases you have issues with? For documentation purposes. Since introduction of contextvars, thread-local data is not something you use often nowadays, so I would like to know your particular use case.

gortibaldik:
Are you referring to the 50% use cases where we use ThreadLocalSingleton ? We used the ThreadLocalSingleton for the database (SQLAlchemy), before we switched to the Singleton. Also for some per-thread variables in our FastAPI app.

Due to the current state of ThreadLocalSingleton (with the finally block) whenever the configuration was wrong the "unbound instance" error was raised. Basically anytime we changed per-customer configuration files the error could have risen up.

tests/unit/providers/singleton/test_thread_local_singleton_py2_py3.py
gortibaldik: Hi, fixed both comments.

ZipFile:
I see. Try experimenting with `ContextLocalSingleton`, this should work better in combination with FastAPI.

Regarding long build times, I do not think there will be any changes soon. It's quite delicate topic (see #487), so we should be careful here.

I'm personally waiting for 2 things:
* [ARM GHA runenrs](https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/) to be available for opensource repos. ARM builds take forever to build right now due to emulation.
* [Cython 3.1](https://github.com/cython/cython/milestone/65). They promised better [Limited API](https://docs.python.org/3/c-api/stable.html#limited-c-api) support, which may significantly reduce amount of packages we build (hence reduce build time).

gortibaldik:
Unfortunately, the project is already in production, and the experimentation is very limited. We likely won't replace ThreadLocalSingleton with ContextLocalSingleton in the forseeable future.

gortibaldik:
Hi, what is the state of this MR ? In my opinion it is really small and improves the functionality of the lib, I don't see any reasons to reject or stall this MR. Can you elaborate on this, please ?

rmk135 (approved):

rmk135:
@gortibaldik thank you for the PR and thanks everyone here for the tons of useful feedback.

rmk135:
All, @ZipFile and I will push a new release shortly.

PR #858, opened 2025-02-21
Feature: Add resource type parameter to init and shutdown resources using specialized providers

amoncusir:
# Objective
Enable the initialization and shutdown of resources by specifying their type. This allows you to create logical groups of resources and handle them by type, similar to applying resource scoping.

# Changes
* New Argument: Added an optional argument to both Container.init_resources and Container.shutdown_resources to specify which resource type(s) should be processed.
* Tests: Added tests to ensure that resource scoping works correctly, including scenarios where dependent resources belong to different types.
* Docs: Added documentation on resource to explain how can be used


colo-o (approved):
Pretty useful!

wadinj:
@ZipFile any changes to make this merge?

PR #898, opened 2025-05-31
Add support for Fast Stream Depends

AndrianEquestrian:
Inspired by  PR [854](https://github.com/ets-labs/python-dependency-injector/pull/854). A slightly cleaner solution, where FastApi separated from Fast Stream properly. Even though they were together at some point, now Fast Stream - is a separate library after all.

### DIFF
--- a/src/dependency_injector/wiring.py
+++ b/src/dependency_injector/wiring.py
@@ -66,6 +66,12 @@ except ImportError:
     werkzeug = None
 
 
+try:
+    import fast_depends.dependencies
+except ImportError:
+    fast_depends = None
+
+
 from . import providers
 
 __all__ = (
@@ -583,6 +589,8 @@ def _unpatch_attribute(patched: PatchedAttribute) -> None:
 
 
 def _extract_marker(parameter: inspect.Parameter) -> Optional["_Marker"]:
+    depends_available = False
+
     if get_origin(parameter.annotation) is Annotated:
         args = get_args(parameter.annotation)
         if len(args) > 1:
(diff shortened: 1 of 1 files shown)


ZipFile (approved):

ZipFile:
Thanks for contribution!

Totorokrut:
With raw Depends it doesn't work in FastStream.
Such test:
```python
@broker.subscriber("hello_world")
@inject
async def on_hello_world(
    input: Input,
    service: Service=Depends(Provide["service"]),
):
    print(f"Service: {service}")
```
Crashes with an error:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for ResponseModel
response
  Input should be a valid dictionary or instance of ResponseModel [type=model_type, input_value=<dependency_injector.wiri...bject at 0x78deeb6cdf70>, input_type=Provide]
```

But with cast=False all works fine:
```python
@broker.subscriber("hello_world")
@inject
async def on_hello_world(
    input: Input,
    service: Service=Depends(Provide["service"], cast=False),
):
    print(f"Service: {service}")
```

AndrianEquestrian:
> With raw Depends it doesn't work in FastStream. Such test:
> 
> ```python
> @broker.subscriber("hello_world")
> @inject
> async def on_hello_world(
>     input: Input,
>     service: Service=Depends(Provide["service"]),
> ):
>     print(f"Service: {service}")
> ```
> 
> Crashes with an error:
> 
> ```
> pydantic_core._pydantic_core.ValidationError: 1 validation error for ResponseModel
> response
>   Input should be a valid dictionary or instance of ResponseModel [type=model_type, input_value=<dependency_injector.wiri...bject at 0x78deeb6cdf70>, input_type=Provide]
> ```
> 
> But with cast=False all works fine:
> 
> ```python
> @broker.subscriber("hello_world")
> @inject
> async def on_hello_world(
>     input: Input,
>     service: Service=Depends(Provide["service"], cast=False),
> ):
>     print(f"Service: {service}")
> ```

Thank You for the observation! Please, include source where you import `broker` decorator from, I think  I should investigate it, do determine the cause of the issue.

Issue #745, opened 2023-09-14
The library can't handle async generators


alexted:
Based on the example [FastAPI + SQLAlchemy example](https://python-dependency-injector.ets-labs.org/examples/fastapi-sqlalchemy.html) from the documentation i made a myself implementation for work with connections of Async SQLA Core. But I came across the fact that your library does not support working with asynchronous generators. Could you please fix it somehow?
Here, here's my code:
```python
from sqlalchemy.ext.asyncio import create_async_engine, AsyncEngine

from src.app.config import AppConfig


class Database:
    def __init__(self, config: AppConfig) -> None:
        self._engine: AsyncEngine = create_async_engine(config.DB_DSN)

    async def get_connection(self)  -> AsyncConnection:
        async with self._engine.begin() as connection:
            yield connection

```

```python
class Container(containers.DeclarativeContainer):
    wiring_config = containers.WiringConfiguration(packages=["src.api"])

    config = providers.Configuration(pydantic_settings=[AppConfig(_env_file=".env")])
    app_cfg = config.get_pydantic_settings()[0]

    logging = providers.Resource(dictConfig, get_logging_config(app_cfg))

    db = providers.Resource(Database, config=app_cfg)

    language_repository = providers.Factory(
        repositories.LanguageRepository,
        connection=db.provided.get_connection.call(),
    )
    ...
```

I expect to get AsyncConnection, when initializing language_repository, but in fact I get `<async_generator object Database.get_connection at 0x7ff63b913c40>` for some reason. It seems that this should not be the case. I have to additionally do await anext(self._connection) in my repositories to get a live sqla connection. I also tried doing `connection=db.provided.provided.get_connection.call().provided.__anext__.call()`, but for some reason I get an inactive sqla connection which is impossible to work with.

colonelpanic8:
I hit this as well, and I think that this is supposed to be a feature. You can avoid it by wrapping with a class or another object that is not a coroutine

colonelpanic8:
> Why do you think this it's not a bug? It's clearly a bug!

it's not. If you read the code you will see that they clearly do this intentionally.

I think the reason is probably something like:

Well if you are exectuting the fetching of a dependency, its not at all clear on which event loop the asynchronous call is supposed to execute. Often, you would want it to execute on the current event loop, but that may not always be the case.

colonelpanic8:
[...]

PR #900, opened 2025-06-02
Add support for async generator injections

### LINKED ISSUE #603: Injecting an async resource into an async generator provides future instead of resource
First of all, thanks a lot for this library - it's really great work you are doing here!

We are relying on this library for a FastAPI project and we noticed that when we try to inject an async resource into an async generator, somehow the injected resource is still wrapped in a future.

We would expect the following minimal example to run without exception, but what we see is

```
AssertionError: resource is of type <class '_asyncio.Future'>
```

Of course we can easily work around that by doing another `await` to get the actual resource, but it would of course be great if this could be fixed.

```python
import asyncio

class Resource:
    pass

async def async_resource():
    yield Resource()

    resource = providers.Resource(async_resource)

@inject
async def async_generator(resource = Provide[Container.resource]):
    yield resource

async def main():
    container = Container()
    container.wire([__name__])
    await container.init_resources()
    generator = async_generator()
    resource = await anext(generator)
    assert isinstance(resource, Resource), f"resource is of type {type(resource)}"

asyncio.run(main())
```

ZipFile:
Fixes #603 and #745.

`_cwiring.pyx` got major overhaul, but otherwise behavior and performance should stay the same.

### DIFF
--- a/src/dependency_injector/_cwiring.pyx
+++ b/src/dependency_injector/_cwiring.pyx
@@ -11,6 +11,7 @@ from .wiring import _Marker
 cimport cython
 
 
+@cython.internal
 @cython.no_gc
 cdef class KWPair:
     cdef str name
--- a/src/dependency_injector/_cwiring.pyx
+++ b/src/dependency_injector/_cwiring.pyx
@@ -38,8 +38,8 @@ cdef class DependencyResolver:
         self.injections = injections
         self.closings = closings
 
-    async def _await_injection(self, p: KWPair, /) -> None:
-        self.to_inject[p.name] = await p.value
+    async def _await_injection(self, kw_pair: KWPair, /) -> None:
+        self.to_inject[kw_pair.name] = await kw_pair.value
 
     cdef object _await_injections(self, to_await: list):
         return gather(*map(self._await_injection, to_await))
--- a/src/dependency_injector/_cwiring.pyx
+++ b/src/dependency_injector/_cwiring.pyx
@@ -1,83 +1,109 @@
 """Wiring optimizations module."""
 
-import asyncio
-import collections.abc
-import inspect
-import types
+from asyncio import gather
+from collections.abc import Awaitable
+from inspect import CO_ITERABLE_COROUTINE
+from types import CoroutineType, GeneratorType
 
+from .providers cimport Provider, Resource, NULL_AWAITABLE
 from .wiring import _Marker
 
-from .providers cimport Provider, Resource
+cimport cython
 
 
-def _sync_inject(object fn, tuple args, dict kwargs, dict injections, dict closings, /):
-    cdef object result
+@cython.no_gc
+cdef class KWPair:
+    cdef str name
+    cdef object value
+
+    def __cinit__(self, str name, object value, /):
+        self.name = name
+        self.value = value
+
+
+cdef inline bint _is_injectable(dict kwargs, str name):
+    return name not in kwargs or isinstance(kwargs[name], _Marker)
+
+
+cdef class DependencyResolver:
+    cdef dict kwargs
     cdef dict to_inject
-    cdef object arg_key
-    cdef Provider provider
+    cdef dict injections
+    cdef dict closings
 
-    to_inject = kwargs.copy()
-    for arg_key, provider in injections.items():
-        if arg_key not in kwargs or isinstance(kwargs[arg_key], _Marker):
-            to_inject[arg_key] = provider()
+    def __init__(self, dict kwargs, dict injections, dict closings, /):
+        self.kwargs = kwargs
+        self.to_inject = kwargs.copy()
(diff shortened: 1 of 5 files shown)


src/dependency_injector/wiring.py
```diff
@@ -1035,36 +1038,42 @@ def is_loader_installed() -> bool:
...
-            fn,
-            args,
-            kwargs,
-            patched.injections,
-            patched.closing,
-        )
+    async def _patched(*args: Any, **raw_kwargs: Any) -> Any:
+        dr = DependencyResolver(raw_kwargs, patched.injections, patched.closing)
+
+        async with dr as kwargs:
```
rmk135, line 1051:
```suggestion
        async with resolver as kwargs:
```

src/dependency_injector/wiring.py
```diff
@@ -1035,36 +1038,42 @@ def is_loader_installed() -> bool:
...
+    return cast(F, _patched)
+
+
+# Async generators too...
+def _get_async_gen_patched(fn: F, patched: PatchedCallable) -> F:
+    @functools.wraps(fn)
+    async def _patched(*args: Any, **raw_kwargs: Any) -> AsyncIterator[Any]:
+        dr = DependencyResolver(raw_kwargs, patched.injections, patched.closing)
+
+        async with dr as kwargs:
```
rmk135, line 1063:
```suggestion
        async with resolver as kwargs:
```

src/dependency_injector/wiring.py
```diff
@@ -1035,36 +1038,42 @@ def is_loader_installed() -> bool:
...
-    def _patched(*args, **kwargs):
-        return _sync_inject(
-            fn,
-            args,
-            kwargs,
-            patched.injections,
-            patched.closing,
-        )
+    def _patched(*args: Any, **raw_kwargs: Any) -> Any:
+        dr = DependencyResolver(raw_kwargs, patched.injections, patched.closing)
```
rmk135, line 1073:
```suggestion
        resolver = DependencyResolver(raw_kwargs, patched.injections, patched.closing)

        with resolver as kwargs:
```

src/dependency_injector/wiring.py
```diff
@@ -1035,36 +1038,42 @@ def is_loader_installed() -> bool:
...
+    async def _patched(*args: Any, **raw_kwargs: Any) -> Any:
+        dr = DependencyResolver(raw_kwargs, patched.injections, patched.closing)
+
+        async with dr as kwargs:
+            return await fn(*args, **kwargs)
+
+    return cast(F, _patched)
+
+
+# Async generators too...
```
rmk135, line 1057:
```suggestion
```

src/dependency_injector/_cwiring.pyx
```diff
@@ -1,83 +1,109 @@
...
+from .providers cimport Provider, Resource, NULL_AWAITABLE
 from .wiring import _Marker
 
-from .providers cimport Provider, Resource
+cimport cython
 
 
-def _sync_inject(object fn, tuple args, dict kwargs, dict injections, dict closings, /):
-    cdef object result
+@cython.no_gc
```
rmk135, line 14: @ZipFile does it only disable the check of circular dependencies in itself or do we need to clean up these objects on our own now?

src/dependency_injector/_cwiring.pyx
```diff
@@ -1,83 +1,109 @@
...
-        if arg_key not in kwargs or isinstance(kwargs[arg_key], _Marker):
-            to_inject[arg_key] = provider()
+    def __init__(self, dict kwargs, dict injections, dict closings, /):
+        self.kwargs = kwargs
+        self.to_inject = kwargs.copy()
+        self.injections = injections
+        self.closings = closings
 
-    result = fn(*args, **to_inject)
+    async def _await_injection(self, p: KWPair, /) -> None:
```
rmk135, line 40:
```suggestion
    async def _await_injection(self, kw_pair: KWPair, /) -> None:
        self.to_inject[kw_pair.name] = await kw_pair.value
```

src/dependency_injector/_cwiring.pyx
ZipFile: Yes, It [disables circular checks](https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#disabling-cyclic-garbage-collection) only (skips `tp_traverse` and `tp_clear` slots). Since this is an internal class, we're confident enough that it'll never reference itself, even indirectly.

Here's full diff with and without `@cython.no_gc` deco: [no_gc.patch](https://github.com/user-attachments/files/20577371/no_gc.patch)

rmk135 (approved):

rmk135:
@ZipFile That's a fantastic improvement! Thank you for doing this!

Issue #926, opened 2025-10-08
providers.Resource is missing overload for AbstractContextManager and AbstractAsyncContextManager


leonarduschen:
```python
# main.py

from collections.abc import Iterator
from contextlib import contextmanager
from typing import reveal_type

from dependency_injector import containers, providers


@contextmanager
def _get_my_int() -> Iterator[int]:
    print("Enter context")
    yield 1


class Container(containers.DeclarativeContainer):
    get_my_int = providers.Resource(_get_my_int)


container = Container()
my_int = container.get_my_int()
reveal_type(my_int)

```

Running `python main.py`:
```
Enter context
Runtime type is 'int'
```

Running `mypy main.py`
```
main.py:20: note: Revealed type is "contextlib._GeneratorContextManager[builtins.int, None, None]"
Success: no issues found in 1 source file
```

leonarduschen:
I think this should do the job (must be placed above the overload for `Callable[..., T]`)

```python
class Resource(Provider[T]):
    .
    .
    .

    @overload
    def __init__(
        self,
        provides: Optional[_Callable[..., contextlib.AbstractContextManager[T]]] = None,
        *args: Injection,
        **kwargs: Injection,
    ) -> None: ...
    @overload
    def __init__(
        self,
        provides: Optional[_Callable[..., contextlib.AbstractAsyncContextManager[T]]] = None,
        *args: Injection,
        **kwargs: Injection,
    ) -> None: ...

    .
    .
    .
```

leonarduschen:
I can open a PR if you haven't worked on it already @ZipFile

ZipFile:
Please do! I haven't started yet.

PR #927, opened 2025-10-10
Fix providers.Resource missing overloads for AbstractContextManager and AbstractAsyncContextManager

leonarduschen:
Addresses #926

---

Demo

```python
# main.py

import asyncio
from collections.abc import AsyncIterator, Iterator
from contextlib import asynccontextmanager, contextmanager
from typing import reveal_type

@contextmanager
def _get_my_int() -> Iterator[int]:
    yield 1


@asynccontextmanager
async def _a_get_my_int() -> AsyncIterator[int]:
    yield 1


    get_my_int = providers.Resource(_get_my_int)
    a_get_my_int = providers.Resource(_a_get_my_int)


async def main() -> None:
    container = Container()
    my_int = container.get_my_int()
    reveal_type(my_int)

    a_my_int = await container.a_get_my_int.async_()
    reveal_type(a_my_int)


if __name__ == "__main__":
    asyncio.run(main())

````

Runtime:
```
python main.py
Runtime type is 'int'
Runtime type is 'int'
```

Mypy output before changing stubs
```
mypy main.py
main.py:29: note: Revealed type is "contextlib._GeneratorContextManager[builtins.int, None, None]"
main.py:32: note: Revealed type is "contextlib._AsyncGeneratorContextManager[builtins.int, None]"
```

Mypy output after changing stubs
```
mypy main.py
main.py:29: note: Revealed type is "builtins.int"
main.py:32: note: Revealed type is "builtins.int"
```

### DIFF
--- a/tests/typing/resource.py
+++ b/tests/typing/resource.py
@@ -1,3 +1,4 @@
+from contextlib import contextmanager, asynccontextmanager
 from typing import (
     Any,
     AsyncGenerator,
@@ -7,6 +8,7 @@ from typing import (
     Iterator,
     List,
     Optional,
+    Self,
 )
 
 from dependency_injector import providers, resources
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/typing/resource.py [hunk 1]
```diff
@@ -109,3 +111,62 @@ async def _provide8() -> None:
 # Test 9: to check string imports
 provider9: providers.Resource[Dict[Any, Any]] = providers.Resource("builtins.dict")
 provider9.set_provides("builtins.dict")
+
+
+# Test 10: to check the return type with classes implementing AbstractContextManager protocol
```
leonarduschen, line 116: Test 10 and 12 (classes implementing `__(a)enter__/__(a)exit__`) were actually already passing because of the first overload `Callable[..., T]`, but I added these tests anyway because they were not covered yet.

ZipFile (approved):

ZipFile:
Thanks!

Issue #728, opened 2023-07-27
Single Database Session Per FastAPI Request Lifecycle


Trinkes:
Hello,
I would like to implement a mechanism that ensures only one database session is created and tied to the FastAPI request lifecycle. The goal is to have a single shared database session across all resources/classes within a request, allowing easy rollback of operations in case of any request-related issues.

Here's the current code example:
```python
import os

from dependency_injector import containers, providers
from dependency_injector.wiring import Provide, inject
from fastapi import FastAPI, Depends
from sqlalchemy import create_engine
from sqlalchemy.orm import declarative_base, sessionmaker
import uvicorn

engine = create_engine("sqlite://")

Base = declarative_base()
Base.metadata.create_all(engine)
SessionLocal = sessionmaker(bind=engine)


class ApplicationContainer(containers.DeclarativeContainer):
    wiring_config = containers.WiringConfiguration(modules=[__name__])
    database: SessionLocal = providers.Factory(SessionLocal)


app = FastAPI()


@app.get("/")
@inject
async def root(
    session_1=Depends(Provide[ApplicationContainer.database]),
    session_2=Depends(Provide[ApplicationContainer.database]),
):
    return {"session_1_id": id(session_1), "session_2_id": id(session_2)}


container = ApplicationContainer()

if __name__ == "__main__":
    uvicorn.run(
        os.path.basename(__file__).replace(".py", "") + ":app",
        host="127.0.0.1",
        port=5000,
        log_level="info",
        reload=True,
    )

```

Currently, when calling the root endpoint, two separate database sessions are created, resulting in different session IDs:
```json
{
  "session_1_id": 4347665504,
  "session_2_id": 4347668912
}
```

However, the desired behavior is to have both arguments (arg and arg2) hold references to the same database session for each request. Therefore, if we call the request again, the ID would change, indicating that a new session was created:
```json
{
  "session_1_id": 4347665504,
  "session_2_id": 4347665504
}
```

The ultimate objective is to achieve a single database session per request, which would simplify the rollback process for any issues that might arise during the request.

Thank you for your attention to this matter, and I look forward to your guidance and suggestions.

dandiep:
I am struggling with this exact same thing - because multiple sessions are created, you can end up with deadlocks. Ideally there would be a Request scope in dependency injector for this type of thing.

jess-hwang:
[...]

Issue #907, opened 2025-07-06
Context Local Scope-Level Resource/Singleton/...


XKTZ:
Hi!

I would like to manage a scope-managed resource with a possible fastapi request. If I write something like:

```py
class MyService:
    def __init__(self, session: Callable[[], Awaitable[int]]) -> None:
        self._session = session

    async def dosth(self):
        print(f"Session ID: {await self._session()}")


"""
I expect
async def sess():
    print("I am starting")
    yield uuid.uuid4().int
    print("I am ending")
"""


async def sess():
    print("I am starting")
    return uuid.uuid4().int


class Container(DeclarativeContainer):
    _sess = providers.Factory(sess)
    _ctx = providers.ContextLocalSingleton(_sess)

    my_service = providers.Resource(MyService, session=_ctx.provider)


container = Container()
app = FastAPI(lifespan=Lifespan(container=container))

@app.get("/")
@inject
async def index(
    service: Annotated[MyService, Depends(Provide[Container.my_service])],
):
    await service.dosth()
    await service.dosth()
    return 1

container.wire(modules=[__name__])

client = TestClient(app)

client.get("/")
client.get("/")

print("I am ok")
```

The code is maximum I would be able to go for now - the desired behaviour is: the two `service.dosth` calls print same ID in one session, but print different ID in two sessions. So without considering the lifetime the thing is already doable. 
However, if we add an extra constraint to the code, which is in the comment section, that we also hope we are able to commit the stuff at the end (with/without a contextmanager, just like waht the `Resource` do), it seems become difficult. Because it seems ContextLocal seems not being able to catch & handle an AsyncContextManager / Generator. The demand of this seems to be general, where the user may need a `yield` a database session every request:

```py
async def get_session(engine):
    async with AsyncSession(engine) as session():
        async with session.begin():
            yield session
```

I have tried a few possible ways:
- Simply use a Factory won't give same ID even in one session, and Factory seems can't catch a correct contextmanager either
- Use Resource can work but Resource is only for global scope, not request level scope. So it can't work nicely in my scenario
- __del__ might also works, but it is fully give the choice to python's GC

May I ask if there are other ways possibly can solve this problem in the scenario? Thanks a lot!

valentino-sm:
@rumbarum 
If a second request comes before the first one has finished, it will share the same session.
`Closing` doesn’t fit ASGI frameworks – I think this should be mentioned in the docs.

[...]

Issue #760, opened 2023-11-01
Custom ResourceFactory provider help


agusmdev:
Hi! I'm trying to handle SQLAlchemy sessions as a resource provider instead of using FastAPI's built-in Depends.

Let me share some code:

The database class:
```python

class Database:
    """Class to handle database connections and sessions"""

    def __init__(self, db_url: str, **kwargs):
        """Initialize the Database class with a database URL and optional keyword arguments"""
        self.db_url = db_url
        self.engine = create_engine(self.db_url, **kwargs, echo=settings.DEBUG_MODE, connect_args={"check_same_thread": False}, pool_size=0)
        self.local_session = sessionmaker(autocommit=False, autoflush=False, bind=self.engine)

    @contextmanager
    def session(self):
        """Context manager to handle sessions"""
        session: Session = self.local_session()
        try:
            yield session
        except Exception as e:
            session.rollback()
        finally:
            session.close()
```
The main container where I create these resources:

```python

def get_db_session(session_factory):
    with session_factory() as db:
        yield db


class AppContainer(containers.DeclarativeContainer):
    """Container to serve all the containers related to the app"""

   # ..... other injections


    # Setup container for talent services
    db = providers.Factory(Database, db_url=settings.DB_URL)

    db_sess = providers.Resource(get_db_session, session_factory=db.provided.session)

    user_container = providers.Container(
        UserContainer,  # and this container has "UserService" that also uses the db session dependency
        db_sess=db_sess,
    )
    # .... 
```

So, I inject the session into the services directly (or in the future in one repository). I don't like injecting the session into the routers directly that's why I prefer not using the FastAPI Depends.

The code provided works as expected, it creates 1 session per request and then closes it, the thing is that this doesn't work in a concurrent environment, I made some local testing with locust, and once I start having many requests executing with multi-threads, the sessions don't work as expected, it seems that the same session is being used in more than 1 request at the same time, so I guess this is something that we could solve if we had something like `FactoryResource` instead of just a Resource that works as a Singleton. I tried to implement this myself by creating a `CustomProvider` but I couldn't make it, so if anyone has any advice I'd appreciate it.

P.S: I checked all the related issues regarding FastAPI and SQLAlchemy sessions in this repo and tried some approaches but none of them worked correctly

theobouwman:
@agusmdev you dont want a session/connection per request right? I thought that what you want is that connections are reused

theobouwman:
[...]

Issue #699, opened 2023-04-25
Resource init/shutdown not called properly with Closing marker.


kkjot88:
Consider this example:
```python
class TestClass:
    @inject
    def test(self, resource=Closing[Provide["resource"]]):
        assert resource == "resource"
        print("test_class")


@inject
def global_test(
    resource=Closing[Provide["resource"]],
    test_class1: TestClass = Provide["test_class"],
    test_class2: TestClass = Provide["test_class"],
):
    assert resource == "resource"
    print("Test start.")
    test_class1.test()
    test_class2.test()
    print("Test end.")


def init_resource():
    print("Init resource. >>")
    yield "resource"
    print("Shutdown resource. <<")


class Container(containers.DeclarativeContainer):
    wiring_config = containers.WiringConfiguration(modules=[__name__])
    resource = providers.Resource(init_resource)
    test_class = providers.Factory(TestClass)


c = Container()

global_test()
```

Result:
```
Init resource. >>
Test start.
test_class
Shutdown resource. <<
Init resource. >>
test_class
Shutdown resource. <<
Test end.
```

To me it seems init_resource() should be called three times, first time when global_test() execution starts (since resource is injected), then second and third time on each of test_class.test() calls. Result shows otherwise. 

I did some debugging and this is what i figured out, see comments:
```cython
cpdef object _provide(self, tuple args, dict kwargs):
    if self.__initialized:	
        return self.__resource
```

```cython
# global_test() exeuction.
def _get_sync_patched(fn, patched: PatchedCallable):
    @functools.wraps(fn)
    def _patched(*args, **kwargs):
        cdef object result
        cdef dict to_inject
        cdef object arg_key
        cdef Provider provider

        to_inject = kwargs.copy()
        for arg_key, provider in patched.injections.items():
            if arg_key not in kwargs or isinstance(kwargs[arg_key], _Marker):
            	# provider.initialized == False, resource gets initialized.
                to_inject[arg_key] = provider()

	# then we go back to global_test() and into test_class1.test() wrapper right away.
        result = fn(*args, **to_inject)

        if patched.closing:
            for arg_key, provider in patched.closing.items():
                if arg_key in kwargs and not isinstance(kwargs[arg_key], _Marker):
                    continue
                if not isinstance(provider, providers.Resource):
                    continue
                # this does nothing since provider.initialized was set to False during
                # the execution of test_class1.test().
                # provider.initialized == False
                provider.shutdown()

        return result
    return _patched


[...]

PR #931, opened 2025-10-22
Add context local resource

elina-israyelyan:
Closes Issue #728 , #760 #907 
Relates to  #699 

Added a `ContextLocalResource` class that inherits from `Resource` and overrides it so that the actual `_resource` and `_shutdowner` are stored and retrieved via `ContextVar`.
This allows each context (such as a FastAPI request) to have its own instance of the resource.
Since FastAPI creates a new context per request, each request now gets its own context-scoped resource.
Because the class extends `Resource`, the `Closing` marker continues to work as expected.

Demo 
Run the following API, then do concurrent requests.  

``` python
from dependency_injector.wiring import Closing, Provide, inject
from fastapi import Depends, FastAPI
from uuid import uuid4

global_list = []


class AsyncSessionLocal:
    def __init__(self):
        self.id = uuid4()

    async def __aenter__(self):
        print("entering session")
        return self

    async def __aexit__(self, exc_type, exc_val, exc_tb):
        print("closing session")

    async def execute(self, user_input):
        return "executing " + user_input


async def get_session() -> AsyncSessionLocal:
    print("start session")
    async with AsyncSessionLocal() as session:
        yield session
    print("close session")


app = FastAPI()


    db = providers.ContextLocalResource(get_session)


@app.get("/")
@inject
async def index(db: AsyncSessionLocal = Depends(Closing[Provide["db"]])):
    global global_list
    if db.id  in global_list:
        raise Exception("The db session is already being used")  # never reaches here
    global_list.append(db.id)
    res = await db.execute("SELECT 1")
    return str(res)


if __name__ == "__main__":
    import uvicorn

    container = Container()
    container.wire(modules=["__main__"])
    uvicorn.run(app, host="localhost", port=8000)
    container.unwire()

```


src/dependency_injector/providers.pyx
```diff
@@ -3867,6 +3867,133 @@ cdef class Resource(Provider):
...
+
+    @property
+    def initialized(self):
+        """Check if resource is initialized."""
+        return self._resource_context_var.get() != self._none
+
+
+    def shutdown(self):
+        """Shutdown resource."""
+        if self._resource_context_var.get() == self._none :
```
ZipFile:
```suggestion
        if not self.initialized:
```

src/dependency_injector/providers.pyx
```diff
@@ -3867,6 +3867,133 @@ cdef class Resource(Provider):
...
+        if self._resource_context_var.get() != self._none:
+            raise Error("Can not copy initialized resource")
+        copied = _memorized_duplicate(self, memo)
+        copied.set_provides(_copy_if_provider(self.provides, memo))
+        copied.set_args(*deepcopy_args(self, self.args, memo))
+        copied.set_kwargs(**deepcopy_kwargs(self, self.kwargs, memo))
+
+        self._copy_overridings(copied, memo)
+
+        return copied
```
ZipFile: It looks like more or less the same version of `__deepcopy__` of the parent class, except initialized check.

Let's update `Resource.__deepcopy__` to use `initialized` property for this check instead. This way We'll avoid duplication.

src/dependency_injector/providers.pyx
```diff
@@ -3867,6 +3867,133 @@ cdef class Resource(Provider):
...
+                resource = await obj.__aenter__()
+                shutdowner = obj.__aexit__
+            elif hasattr(obj, '__enter__') and hasattr(obj, '__exit__'):
+                resource = obj.__enter__()
+                shutdowner = obj.__exit__
+            else:
+                resource = obj
+                shutdowner = None
+
+            return resource, shutdowner
```
ZipFile: Let's move this into util function and reuse in both versions of resource provider.

src/dependency_injector/providers.pyx
```diff
@@ -3867,6 +3867,133 @@ cdef class Resource(Provider):
...
+        except:
+            raise
```
ZipFile: Seems extra.
```suggestion
```

src/dependency_injector/providers.pyx
```diff
@@ -3867,6 +3867,133 @@ cdef class Resource(Provider):
...
+        if self._async_mode == ASYNC_MODE_ENABLED:
+            return NULL_AWAITABLE
+
+    def _reset_all_contex_vars(self):
+        self._resource_context_var.set(self._none)
+        self._shutdowner_context_var.set(self._none)
+
+
+    async def _shutdown_async(self, future) -> None:
+        await future
```
ZipFile: This is more or less equivalent of:

```python
    def _shutdown_async(self, future) -> Awaitable[Any]:
        return future
```

Which does not make much sense. Lets remove it.

```suggestion
```

src/dependency_injector/providers.pyx
```diff
@@ -3867,6 +3867,133 @@ cdef class Resource(Provider):
...
+        elif hasattr(obj, '__aenter__') and hasattr(obj, '__aexit__'):
+            resource = ensure_future(self._handle_async_cm(obj))
+            self._resource_context_var.set(resource)
+            self._shutdowner_context_var.set(obj.__aexit__)
+            return resource
+        else:
+            self._resource_context_var.set(obj)
+            self._shutdowner_context_var.set(self._none)
+
+        return self._resource_context_var.get()
```
ZipFile: This also looks suspiciously duplicated piece of code. I think we can use custom properties (getters/setters) here instead, to avoid duplication. In plain `Resource` they'll just set private attribute, but in a `ContextLocalResource` - utilize contextvars.

src/dependency_injector/providers.pyx
```diff
@@ -3867,6 +3867,133 @@ cdef class Resource(Provider):
...
+        try:
+            resource, shutdowner = result.result()
+        except Exception as exception:
+            self._resource_context_var.set(self._none)
+            self._shutdowner_context_var.set(self._none)
+            future_result.set_exception(exception)
+        else:
+            self._resource_context_var.set(resource)
+            self._shutdowner_context_var.set(shutdowner)
+            future_result.set_result(resource)
```
ZipFile: See suggestion above about properties.

src/dependency_injector/providers.pyx
```diff
@@ -3867,6 +3867,133 @@ cdef class Resource(Provider):
         return self._resource
 
 
+cdef class ContextLocalResource(Resource):
```
ZipFile: As I've already mentioned, I think we'll need `BaseResource` with common reusable code. Do not hesitate to edit existing code.

ZipFile (changes requested):
Hey! Thanks for contribution. Idea is good, but there are some code duplication issues needs to be addressed before we can merge it. Also, it would be really nice to have documentation updated with usage examples.

src/dependency_injector/providers.pyx
elina-israyelyan: moved it to Resource, used the same logic in both places

src/dependency_injector/providers.pyx
elina-israyelyan: Used Resource as Base class. Didnt make BaseResource as if we do it then it would make sense to rename Resource -> SingletonResource which I avoided for not changing any interface

examples/providers/context_local_resource.py
```diff
@@ -0,0 +1,50 @@
+from uuid import uuid4
+
+from fastapi import Depends, FastAPI
+
+from dependency_injector import containers, providers
+from dependency_injector.wiring import Closing, Provide, inject
+
+global_list = []
```
ZipFile, line 8: I think we can get rid of this list. We never print/log/return its contents anywhere, so for the sake of example it looks useless a bit.

examples/providers/context_local_resource.py
```diff
@@ -0,0 +1,50 @@
...
+app = FastAPI()
+
+
+class Container(containers.DeclarativeContainer):
+    db_session = providers.ContextLocalResource(AsyncSessionLocal)
+
+
+@app.get("/")
+@inject
+async def index(db: AsyncSessionLocal = Depends(Closing[Provide["db_session"]])):
```
ZipFile, line 35: I do not think this fully demonstrates "context local" part of the `ContextLocalResource`. With only one injection it looks as if it was a regular `Resource`. Let's add few more injections to the mix (e.g. "auth" middleware, or something) to demonstrate that we're, in fact, getting the same instance of `AsyncSessionLocal` within the same request context.

elina-israyelyan:
> Hey! Thanks for contribution. Idea is good, but there are some code duplication issues needs to be addressed before we can merge it. Also, it would be really nice to have documentation updated with usage examples.

Hey @ZipFile , I resloved all the comments and added a small reference in the documentation with small example.

elina-israyelyan:
Hey @ZipFile , I resloved all the comments and added a small reference in the documentation with small example.

ZipFile (commented):
Hey! I'm still aware of this PR. It's rather busy end of the year for me, so not much time to take care of it.

I have few concerns regarding of the C code Cython generates here, so I'm experimenting a bit with a different ways to approach it (on my rather limited free time). Also I feel like we can reduce amount of cvars used here. But, unfortunately, no concrete recommendations how can we improve it atm.

In the mean time, there few comments regarding the documentation/examples.

examples/providers/context_local_resource.py
elina-israyelyan: Do you mean having nested injections ? I am a bit confused about why only that doesnt show the context "locality"

elina-israyelyan: Or you mean adding Resource to showcase the difference between ContextLocalResource and Resource

ZipFile: > Do you mean having nested injections ?

This. `@inject` same resource in a different place to show that both receive the same instance of `AsyncSessionLocal`.

elina-israyelyan: added

src/dependency_injector/providers.pyx
```diff
@@ -3866,6 +3890,57 @@ cdef class Resource(Provider):
...
+    def __init__(self, provides=None, *args, **kwargs):
+        self._initialized_context_var = ContextVar("_initialized_context_var", default=False)
+        self._resource_context_var = ContextVar("_resource_context_var", default=None)
+        self._shutdowner_context_var = ContextVar("_shutdowner_context_var", default=None)
+        super().__init__(provides, *args, **kwargs)
+
+    @property
+    def _initialized(self):
+        """Get initialized state."""
+        return self._initialized_context_var.get()
```
gregbrowndev: as per [your comment](https://github.com/ets-labs/python-dependency-injector/pull/931#pullrequestreview-3408775101) here @ZipFile, maybe you can just use 1 ContextVar here with a tuple/frozen dataclass for the data? (e.g. to avoid sharing a mutable default dict)

```py
cdef class ContextLocalResource(Resource):
    def __init__(self, provides=None, *args, **kwargs):
        self._context_var = ContextVar("_context_var", default=(False, None, None))
        super().__init__(provides, *args, **kwargs)

    @property
    def _initialized(self):
        """Get initialized state."""
        return self._context_var.get()[0]

    @_initialized.setter
    def _initialized(self, value):
        """Set initialized state."""
        current = self._context_var.get()
        self._context_var.set((value, current[1], current[2]))

    @property
    def _resource(self):
        """Get resource."""
        return self._context_var.get()[1]

    @_resource.setter
    def _resource(self, value):
        """Set resource."""
        current = self._context_var.get()
        self._context_var.set((current[0], value, current[2]))

    @property
    def _shutdowner(self):
        """Get shutdowner."""
        return self._context_var.get()[2]

    @_shutdowner.setter
    def _shutdowner(self, value):
        """Set shutdowner."""
        current = self._context_var.get()
        self._context_var.set((current[0], current[1], value))
```

ZipFile (approved):

ZipFile:
Thanks! Available in 4.49.0.

Issue #930, opened 2025-10-21
Implement a way to specify typevar of Selector


leonarduschen:
## Background

My usual usecase of Selector is when I have an interface with multiple implementations:

```python
class Animal:
    def __init__(self) -> None:
        pass


class Dog(Animal):
    pass


class Cat(Animal):
    pass
```

And then the container:

```python
class Container(containers.DeclarativeContainer):
    get_dog = providers.Singleton(Dog)
    get_cat = providers.Singleton(Cat)
    get_animal = providers.Selector(lambda: random.choice(["cat", "dog"]), cat=get_cat, dog=get_dog)
```

Usage:

```python
def main() -> None:
    container = Container()

    animal = container.get_animal()
    reveal_type(animal)  # Revealed type is Any
```

---

## Desired behaviors

 
Ideally, `Selector` should:
1. Allow users to specify type: `get_animal = providers.Selector[Animal](lambda: random.choice(["cat", "dog"]), cat=get_cat, dog=get_dog)`
2. Automatically infer from the kwargs of providers: cat and dog are both animals, so get_animal is `Selector[Animal]`

leonarduschen:
## Possible solutions I have explored

TL;DR:

- Solution 1: gives desired behavior 1, but forces users to manually-type
- Solution 2: gives desired behavior 1, but only works on Python 3.13 or above
- Solution 3: gives both desired behavior 1 and 2, but requires changing `Provider` to be covariant

---

1. **Simply making `Selector` generic to `T = TypeVar("T")`**

    Stub
    ```python
    class Selector(Provider[T]):
        def __init__(
            self, selector: Optional[_Callable[..., Any]] = None, **providers: Provider
        ): ...
    ```

    Result
    ```python
    import random
    from typing import reveal_type

    from dependency_injector import containers, providers


    class Animal:
        def __init__(self) -> None:
            pass


    class Dog(Animal):
        pass


    class Cat(Animal):
        pass


    class Container(containers.DeclarativeContainer):
        get_dog = providers.Singleton(Dog)
        get_cat = providers.Singleton(Cat)

        # Not providing type results in error: 'Need type annotation for "get_animal"  [var-annotated]'
        get_animal = providers.Selector(lambda: random.choice(["cat", "dog"]), cat=get_cat, dog=get_dog)
        get_animal_with_type = providers.Selector[Animal](lambda: random.choice(["cat", "dog"]), cat=get_cat, dog=get_dog)


    def main() -> None:
[...]

PR #932, opened 2025-10-23
Allow explicit typing on selector using TypeVar with default Any

leonarduschen:
Partially addresses #930


ZipFile:
Could you rebase with latest `develop` branch? Failing tests should be fixed now (#933).

src/dependency_injector/providers.pyi
```diff
@@ -22,6 +21,8 @@ from typing import (
     overload,
 )
 
+from typing_extensions import TypeVar
```
ZipFile: We need to bump minimal Python version for `typing_extensions` [here](https://github.com/ets-labs/python-dependency-injector/blob/059f78b27ccecabc5f8cf68f181e3253002c8f34/pyproject.toml#L58).

src/dependency_injector/providers.pyi
```diff
@@ -542,17 +544,17 @@ class Container(Provider[T]):
...
     ): ...
     def __getattr__(self, name: str) -> Provider: ...
     @property
     def selector(self) -> Optional[_Callable[..., Any]]: ...
-    def set_selector(self, selector: Optional[_Callable[..., Any]]) -> Selector: ...
+    def set_selector(self, selector: Optional[_Callable[..., Any]]) -> Selector[T_Any]: ...
     @property
     def providers(self) -> _Dict[str, Provider]: ...
-    def set_providers(self, **providers: Provider) -> Selector: ...
+    def set_providers(self, **providers: Provider) -> Selector[TT]: ...
```
ZipFile:
```suggestion
    def set_providers(self, **providers: Provider) -> Self: ...
```
(`Self` from `typing_extensions`)

src/dependency_injector/providers.pyi
leonarduschen: Right, bumped to v3.13 for TypeVar default

src/dependency_injector/providers.pyi
leonarduschen: Ah, I think I wanted to use `T_Any` on this one (was going to check everything before converting the PR from "draft" to "ready")

---

Anyway, changed to `Selector[T_Any]` since similar usecases in this module are not using `typing_extensions.Self` either, let me know if you want me to use typing_extensions.Self, the difference will only be when subclassing `Selector`.

leonarduschen: Note:

We'll have a decision to make on `set_providers` later once we make Provider covariant


**Option 1:**

Require `Selector[MyType].set_provider` to only accept the providers assignable to `Provider[MyType]`, and so we don't change the returned Selector type

```python
    def set_providers(self, **providers: Provider[T_Any]) -> Selector[T_Any]: ...
```

**Option 2:**

Allow any kind of providers and use union for the return type

```python
    def set_providers(self, **providers: Provider[TT_Any]) -> Selector[T_Any | TT_Any]: ...
```

---

In this PR we make it `    def set_providers(self, **providers: Provider) -> Selector[T_Any]: ...` (not changing the retuned Selector type) so we're already leaning toward the first option. I personally prefer it this way to, easier to make sense of

leonarduschen:
Thanks for the quick review!

I've rebased and addressed your comments @ZipFile.

leonarduschen:
Sorry, just added more changes:
- Specify return type of `Selector.__getattr__` and `Selector.providers`
- Added more tests to make sure untyped Selector is not affected by these changes

src/dependency_injector/providers.pyi
ZipFile: > Anyway, changed to `Selector[T_Any]` since similar usecases in this module are not using typing_extensions.Self either, let me know if you want me to use `typing_extensions.Self`, the difference will only be when subclassing `Selector`.

I was thinking about `typing_extensions.Self` from more practical standpoint. This method literally [returns self](https://github.com/ets-labs/python-dependency-injector/blob/5a1aef920372da6a95d14865caab10061aea7104/src/dependency_injector/providers.pyx#L4107-L4110). This should save us some headache in the future.

> Require Selector[MyType].set_provider to only accept the providers assignable to Provider[MyType], and so we don't change the returned Selector type

Looks reasonable enough.

```python
from typing_extensions import Self as _Self  # alias to not collide with Self provider

def set_providers(self, **providers: Provider[T_Any]) -> _Self: ...
```

leonarduschen: Sure thing

ZipFile (approved):

ZipFile:
Thanks!

Issue #934, opened 2025-10-24
Use assert_type for type-stub checks intended for exact type match


leonarduschen:
Checks in `tests/typing` intended to be exact matches should be done with `assert_type` instead of simply assigning to variables

## Explanation

Consider the following functions
```python
def f_any() -> Any:
    return ...

def f_int() -> int:
    return 1
```

Assigning to vars to check
```python
# It might look like we're asserting the return type of function is Any
var1: Any = f_any()

# But all it does it asserting that return_any() is assignable to Any 
# which is always true for any functions

var2: Any = f_int() # This is ok with mypy
```

Using assert_type
```python
assert_type(f_any(), Any)
assert_type(f_int(), Any)  # error: Expression is of type "int", not "Any"  [assert-type]
```

---

The usecase for simply assigning to variables is to check if a variable is a subclass (or "assignable" to be specific) of a specific type. For example.

```python
var1: Animal = f_cat()  # We don't care if the stub declares f_cat to return Animal or Cat
```

But in the usecase of stub-testing, we probably almost always want to check exact type insteaad of assignability.

## Examples

**Example 1:**
https://github.com/ets-labs/python-dependency-injector/blob/5a1aef920372da6a95d14865caab10061aea7104/tests/typing/selector.py#L11

This will pass regardless of the return type of `provider1`

**Example 2:**
https://github.com/ets-labs/python-dependency-injector/blob/5a1aef920372da6a95d14865caab10061aea7104/tests/typing/factory.py#L92

The example is supposed to check if explicitly typing  `factory10: Provider.Animal` makes `factory10()` return `Animal` instead of `Cat`.

But it doesn't check that, since `Cat` is assignable to `Animal`, which defeats the purpose of the test

```python
factory10 = providers.Factory(Cat)
animal10: Animal = factory10()  # Here will still pass even though it returns cat
```

This, on the other hand, will work as intended
```python
factory10: providers.Provider[Animal] = providers.Factory(Cat)
assert_type(factory10(), Animal)
```

---

There are probably a lot more examples

ZipFile:
`assert_type` is a Python 3.11 feature, those test were written before this function even existed. If you're willing to retrofit them, I'll gladly accept a PR.

leonarduschen:
Right, I noticed the issue when working on the Selector typevar, I thought I'd open a new issue to discuss first to make sure I understand the problem correctly :)

[...]

PR #935, opened 2025-10-25
Retrofit assert_type for Aggregate and Callable

leonarduschen:
Partially addresses #934

### DIFF
--- a/tests/typing/aggregate.py
+++ b/tests/typing/aggregate.py
@@ -1,4 +1,5 @@
 from dependency_injector import providers
+from typing_extensions import assert_type, Any
 
 
 class Animal: ...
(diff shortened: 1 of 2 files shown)


tests/typing/aggregate.py
```diff
@@ -8,29 +9,36 @@ class Cat(Animal): ...
...
 provider1_set_non_string_keys.set_providers({Cat: providers.Object("str")})
-provider_set_non_string_1: providers.Provider[str] = (
-    provider1_set_non_string_keys.providers[Cat]
-)
+provider_set_non_string_1 = provider1_set_non_string_keys.providers[Cat]
+assert_type(provider_set_non_string_1, providers.Provider[str])
+
 
-provider1_new_non_string_keys: providers.Aggregate[str] = providers.Aggregate(
+# TODO: Change providers.Aggregate to accept Mapping? Then remove explicit typing here
```
leonarduschen: Current stub:

```python
class Aggregate(Provider[T]):
    def __init__(
        self,
        provider_dict: Optional[_Dict[Any, Provider[T]]] = None,
        **provider_kwargs: Provider[T],
    ): ...
```

I think we should accept `collections.abc.Mapping` here

```python
class Aggregate(Provider[T]):
    def __init__(
        self,
        provider_dict: Optional[Mapping[Any, Provider[T]]] = None,
        **provider_kwargs: Provider[T],
    ): ...
```

Because `Dict` is invariant in both key and value, so `providers.Aggregate` cannot infer its own typevar without explicit typing

```python
provider1_new_non_string_keys = providers.Aggregate(
    {Cat: providers.Object("str")},
)  # This raises mypy error
```

tests/typing/callable.py
```diff
@@ -28,24 +30,34 @@ def create(cls) -> Animal:
 
 # Test 4: to check the .args & .kwargs attributes
 provider4 = providers.Callable(Animal)
-args4: Tuple[Any] = provider4.args
-kwargs4: Dict[str, Any] = provider4.kwargs
+args4 = provider4.args
+kwargs4 = provider4.kwargs
+assert_type(args4, Tuple[Any])
+# TODO: Change Callable.kwargs to Dict[str, Any]? Then adjust test back to Dict[str, Any]
```
leonarduschen: The stub actually declares `provider4.kwargs` to return `Dict[Any, Any]`, not `Dict[str, Any]` like the previous test probably intended to check.

But I think we can safely change it to return `Dict[str, Any]` in the stub.

tests/typing/callable.py
```diff
@@ -28,24 +30,34 @@ def create(cls) -> Animal:
...
 provider5 = providers.Callable(Animal)
-provided5: Animal = provider5.provided()
-attr_getter5: providers.AttributeGetter = provider5.provided.attr
-item_getter5: providers.ItemGetter = provider5.provided["item"]
-method_caller: providers.MethodCaller = provider5.provided.method.call(123, arg=324)
+provided_val5 = provider5.provided()
+attr_getter5 = provider5.provided.attr
+item_getter5 = provider5.provided["item"]
+method_caller5 = provider5.provided.method.call(123, arg=324)
+# TODO: Remove explicit typing of Provider.provided return type
```
leonarduschen: The stub is this

```python
    @property
    def provided(self) -> ProvidedInstance[T]: ...
```

But` ProvidedInstance` is not declared to be generic

```python
class ProvidedInstance(Provider, ProvidedInstanceFluentInterface):
    def __init__(self, provides: Optional[Provider] = None) -> None: ...
```

I think we can make the quick fix to
```python
    @property
    def provided(self) -> ProvidedInstance: ...
```

tests/typing/callable.py
```diff
@@ -55,20 +67,23 @@ def create(cls) -> Animal:
...
-    animal2: Animal = await provider9.async_(1, 2, 3, b="1", c=2, e=0.0)
+    await provider9(1, 2, 3, b="1", c=2, e=0.0)  # type: ignore[misc]
+    cat9 = await provider9.async_(1, 2, 3, b="1", c=2, e=0.0)
+    assert_type(cat9, Cat)
 
 
 # Test 10: to check the .provides
 provider10 = providers.Callable(Cat)
-provides10: Optional[Callable[..., Cat]] = provider10.provides
-assert provides10 is Cat
```
leonarduschen, line 65: I think these assert statements are not doing anything

tests/typing/callable.py
```diff
@@ -55,20 +67,23 @@ def create(cls) -> Animal:
...
 provider11 = providers.Callable[Animal](Cat)
-provides11: Optional[Callable[..., Animal]] = provider11.provides
-assert provides11 is Cat
+provides11 = provider11.provides
+assert_type(provides11, Optional[Callable[..., Animal]])
+
 
 # Test 12: to check string imports
-provider12: providers.Callable[Dict[Any, Any]] = providers.Callable("builtins.dict")
+# TODO: Use T_Any as Callable typevar? Then remove type-ignore
```
leonarduschen: Can't infer typevar without explicit typing, mypy raises error here, but we can change it to typevar with default Any to fix it

leonarduschen (commented):
I've only worked on aggregate.py and callable.py, but found some areas that need quick-fixings.

I think it'd make more sense to include these quick-fixes in the same PR, as the tests are supposed to check exactly the behaviors these quick-fixes are supposed to affect.

@ZipFile let me know what you think.

ZipFile (commented):
Hello! Sorry for the late review. I've addressed issues you've mentioned. Can you rebase the branch with latest `develop` and remove `TODO`s?

leonarduschen:
Sure!

I'll do it sometime this week.

ZipFile (approved):

PR #933, opened 2025-10-23
Fast depends v3 compatibility fix

AndrianEquestrian:
It seems that after the most recent release of Fast Depends, our integration and corresponding tests were broken. This branch contains fix: parse new model - Dependant, for Fast Depends of version 3

PR #939, opened 2025-12-04
Allow `Annotated` marker to be anywhere in the annotation list.

BrianPugh:
There are some situations where multiple libraries/frameworks are inspecting `Annotated` types. For example, the following example code using the [Cyclopts](https://github.com/BrianPugh/cyclopts) CLI library:

```python
from cyclopts import Parameter

@app.default
@inject
def main(
    name: str,
    *,
    greeting_service: Annotated[GreetingService, Provide[Container.greeting_service], Parameter(parse=False)],
):
    ...
```

Currently, `dependency_injector` [is only looking at the first annotation](https://github.com/ets-labs/python-dependency-injector/blob/5a1aef920372da6a95d14865caab10061aea7104/src/dependency_injector/wiring.py#L677). This PR updates the logic to use the **first valid extracted marker**. This way, the `dependency_injector` marker no longer has to be the first annotated element.

Issue that brought this up: https://github.com/BrianPugh/cyclopts/issues/704

### DIFF
--- a/src/dependency_injector/wiring.py
+++ b/src/dependency_injector/wiring.py
@@ -682,23 +682,18 @@ def _unpatch_attribute(patched: PatchedAttribute) -> None:
 
 def _extract_marker(parameter: inspect.Parameter) -> Optional["_Marker"]:
     if get_origin(parameter.annotation) is Annotated:
-        args = get_args(parameter.annotation)
-        if len(args) > 1:
-            marker = args[1]
-        else:
-            marker = None
+        candidates = get_args(parameter.annotation)[1:]
     else:
-        marker = parameter.default
+        candidates = (parameter.default,)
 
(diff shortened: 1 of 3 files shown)


BrianPugh:
#940 needs to be merged in first to fix tests.

ZipFile:
Thanks! Fix from #940 is not really needed, since we already had one in #933.

ZipFile:
Available in v4.48.3

jochumdev:
For anyone looking into integrating those 2:

https://github.com/BrianPugh/cyclopts/issues/704#issuecomment-3613949704

@BrianPugh this was bugging me too, thx for fixing it!

Edit: I also love to use python-dependency-injector, it's a very very powerful tool.

PR #940, opened 2025-12-04
FastDepends v3 compatibility

BrianPugh:
Currently tests in this repo are failing. This is because [FastDepends v3 was released in mid-october](https://github.com/Lancetnik/FastDepends/releases/tag/3.0.0) and had a slight API change. This PR adds support in a v2-backwards-compatible way.


ZipFile:
See https://github.com/ets-labs/python-dependency-injector/pull/939#issuecomment-3613668568

Issue #945, opened 2025-12-10
Add provided()<func>.call *args, **kwargs arguments


pavalso:
The title is self-explanatory. This feature would make it possible to pass arguments to the `call` function, which currently lacks this capability. I believe this would be a valuable addition.

I give some examples that would work with this feature:

Being logger a providers.Singleton(Logger, "base")
``` python
Provide["logger", provided().getChild.call("child")]
# This once wired would resolve into base.child logger
```

ZipFile:
Available in 4.49.0

PR #946, opened 2025-12-10
Add provided()<func>.call *args, **kwargs arguments #945

pavalso:
#945

### DIFF
--- a/tests/unit/samples/wiring/module.py
+++ b/tests/unit/samples/wiring/module.py
@@ -101,6 +101,27 @@ def test_provided_instance(some_value: int = Provide[Container.service.provided.
 
 
 @inject
+def test_provided_instance_call_with_args(
+        some_value: int = Provide[Container.service_with_callable.provided.method_with_args.call(1, 2)]
+):
+    return some_value
+
+
+@inject
+def test_provided_instance_call_with_kwargs(
+        some_value: dict = Provide[Container.service_with_callable.provided.method_with_kwargs.call(a=1, b=2)]
+):
+    return some_value
+
+
+@inject
+def test_provided_instance_call_with_args_and_kwargs(
+        some_value: dict = Provide[Container.service_with_callable.provided.foo.process.call(1, 2, key="value")]
+):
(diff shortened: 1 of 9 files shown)


ZipFile (changes requested):
Looks fine, but we need test coverage for this new feature.

tests/unit/samples/wiring/module.py
```diff
@@ -100,6 +100,27 @@ def test_provided_instance(some_value: int = Provide[Container.service.provided.
     return some_value
 
 
+@inject
+def test_provided_instance_call_with_args(
+        some_value: int = Provide[Container.service_with_callable.provided.method_with_args.call(1, 2)]
```
ZipFile, line 105: Shouldn't we test for string ids? I.e. `Provide["service_with_callable", provided().method_with_args.call(1, 2)]`?

tests/unit/samples/wiring/service.py
```diff
@@ -1,2 +1,15 @@
 class Service:
     service_attr: int
+
+
+class ServiceWithCallable:
+    def __init__(self):
+        self.foo = CallableDict({"bar": lambda: 10})
+        self.method_with_args = lambda x, y: x + y
+        self.method_with_kwargs = lambda **kwargs: kwargs
```
ZipFile: I think it would be better to have them as proper methods

tests/unit/samples/wiring/service.py
```diff
@@ -1,2 +1,15 @@
 class Service:
     service_attr: int
+
+
+class ServiceWithCallable:
+    def __init__(self):
+        self.foo = CallableDict({"bar": lambda: 10})
```
ZipFile, line 7:
```suggestion
        self.foo = {"bar": lambda *args, **kwargs: {"args": args, "kwargs": kwargs}}
```
Does it need to be separate dict subclass?

ZipFile (approved):

ZipFile:
Thanks! Available in 4.49.0.

PR #943, opened 2025-12-08
Retrofit assert type for some type-stub checks

leonarduschen:
Partially addresses #934 

Continuation of #935

Retrofitted for:
    - configuration.py
    - container.py
    - declarative_container.py
    - delegate.py
    - dependencies_container.py
    - dependency.py
    - dict.py
    - dynamic_container.py

PR #953, opened 2026-02-08
Retrofit assert type for remaining type-stub checks

### LINKED ISSUE #934: Use assert_type for type-stub checks intended for exact type match

leonarduschen:
Closes #934

Continuation of #935 and #943

Retrofitted for all remaining type-stub checks:

- factory.py
- list.py
- object.py
- provider.py
- resource.py
- singleton.py


leonarduschen:
@ZipFile last PR for #934, ready for review

ZipFile (approved):

ZipFile:
Thanks! Available in 4.49.0.

PR #958, opened 2026-04-17
Add FastStream example to docs

birthdaysgift:
Despite `FastStream` uses `FastDepends` under the hood the existing example for `FastDepends` from [documentation](https://python-dependency-injector.ets-labs.org/examples/fastdepends.html#) doesn't work out of the box when used in `FastStream` handlers.

So I thought that having an explicit example of how to use `dependency_injector` with `FastStream` would be helpful for other devs seeking for a solution for such integration.


docs/examples/faststream.rst
```diff
@@ -0,0 +1,42 @@
...
+
+FastStream example
+=============
+
+.. meta::
+   :keywords: Python,Dependency Injection,FastStream,Example
+   :description: This example demonstrates a usage of FastStream with Dependency Injector.
+
+
+This example shows how to use ``Dependency Injector`` with `FastStream <https://dummy.faststream.airt.ai/0.5/faststream/>`_.
```
ZipFile, line 11:
```suggestion
This example shows how to use ``Dependency Injector`` with `FastStream <https://github.com/ag2ai/faststream>`_.
```

docs/examples/faststream.rst
```diff
@@ -0,0 +1,42 @@
+.. _faststream-example:
+
+FastStream example
+=============
```
ZipFile, line 4:
```suggestion
FastStream example
==================
```

examples/miniapps/faststream/src/consumer.py
```diff
@@ -0,0 +1,57 @@
...
+    def next(self) -> int:
+        self.count += 1
+        return self.count
+
+
+class Container(containers.DeclarativeContainer):
+    counter = providers.Singleton(Counter)
+
+
+broker = RedisBroker("redis://redis", logger=None)
```
ZipFile, line 24:
```suggestion
    config = providers.Configuration()
    counter = providers.Singleton(Counter)
    broker = providers.Resource(RedisBroker, config.redis_url, logger=None)
    app = providers.Factory(FastStream, broker, logger=None)
```

If we're demonstrating integration, it would be better to go all-out.

examples/miniapps/faststream/src/consumer.py
```diff
@@ -0,0 +1,57 @@
...
+) -> None:
+    count = counter.next()
+    print(f"Message #{count} from {message.user}: '{message.text}'")
+
+
+async def main() -> None:
+    container = Container()
+    container.wire(modules=[__name__])
+
+    await FastStream(broker, logger=None).run()
```
ZipFile, line 52: see comment above

```suggestion
    app = await container.app()

    await app.run()
```

examples/miniapps/faststream/src/producer.py
```diff
@@ -0,0 +1,25 @@
+import json
+import time
+
+import redis
+
+
+def main():
+    client = redis.Redis(host="redis", port=6379)
```
ZipFile, line 8: No DI container?

examples/miniapps/faststream/Dockerfile.producer
ZipFile, line 1: For simplicity one Dockerfile should suffice. Set different entrypoints in `docker-compose.yml`.

examples/miniapps/faststream/Dockerfile.consumer
```diff
@@ -0,0 +1,11 @@
...
+
+WORKDIR /app
+
+COPY requirements.txt ./
+RUN pip install -r requirements.txt
+
+COPY ./src ./src
+
+ENV PYTHONUNBUFFERED=1
+CMD ["python3", "-m", "src.consumer"]
```
ZipFile, line 11: For [src layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/), you have to do [few extra steps](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html) in `pyproject.toml`. Either configure it properly (`src` should not be an importable package) or just use flat layout (without `src/`).

examples/miniapps/faststream/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+name: faststream-example
+
+services:
+
+  redis:
+    container_name: "${COMPOSE_PROJECT_NAME}-redis"
```
ZipFile, line 6:
```suggestion
```
IIRC this is a default docker compose behavior, so this is extra.

examples/miniapps/faststream/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+    image: redis
+
+  consumer:
+    container_name: "${COMPOSE_PROJECT_NAME}-consumer"
```
ZipFile, line 10:
```suggestion
```
See comment above.

examples/miniapps/faststream/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+    build:
+      dockerfile: Dockerfile.consumer
+    depends_on:
+      - redis
+
+  producer:
+    container_name: "${COMPOSE_PROJECT_NAME}-producer"
```
ZipFile, line 17:
```suggestion
```
See comment above.

examples/miniapps/faststream/README.rst
```diff
@@ -0,0 +1,40 @@
+FastStream + Dependency Injector Example
+=====================================
```
ZipFile, line 2:
```suggestion
FastStream + Dependency Injector Example
========================================
```

examples/miniapps/faststream/README.rst
```diff
@@ -0,0 +1,40 @@
...
+Everything can be run via docker compose.
+
+A convenient ``run.sh`` script runs consumer, producer and redis services, prints logs from consumer
+and shuts down once producer exits.
+
+Ensure that ``run.sh`` has execution permission:
+
+.. code-block:: bash
+
+    sudo chmod +x ./run.sh
```
ZipFile, line 23:
```suggestion
```
You can `chmod +x ./run.sh` and commit it right away. Git keeps track of mode flags.

examples/miniapps/faststream/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+    build:
+      dockerfile: Dockerfile.producer
```
ZipFile, line 19:
```suggestion
  consumer:
    image: faststream-example
    depends_on:
      - redis

  producer:
    image: faststream-example
```
See comment about extra Dockerfile.

docs/examples/faststream.rst
ZipFile, line 1: Add reference to `docs/examples/index.rst` so it appears alongside other examples.

examples/miniapps/faststream/run.sh
```diff
@@ -0,0 +1,15 @@
...
+    --exit-code-from producer
+
+docker container rm \
+    faststream-example-producer \
+    faststream-example-consumer \
+    faststream-example-redis
+
+docker image rm \
+    faststream-example-producer \
+    faststream-example-consumer
```
ZipFile, line 15:
```suggestion
docker compose down --rmi local
```
Though, I'm not sure if this is a desired behavior, since user might want to tinker with an example app for longer. Better include it in the documentation.

examples/miniapps/faststream/run.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
```
ZipFile, line 2:
```suggestion
set +x
```

ZipFile (changes requested):

examples/miniapps/faststream/docker-compose.yml
birthdaysgift: nope, by default it's `faststream-example-redis-1` instead of `faststream-example-redis`

I prefer to have container names explicit because I assume that one day default docker compose behavior might be changed so all my scripts that rely on those container names will stop working

examples/miniapps/faststream/Dockerfile.consumer
birthdaysgift: I think this suggestion relates to installable python packages (i.e. libraries). Since my example is just a runnable application I don't see why it should follow this approach.

Does it make a sense for runnable apps?

examples/miniapps/faststream/docker-compose.yml
ZipFile: From the code you've submitted it looks like the only reason to have explicit names is to do `docker container rm` later. With docker compose you can accomplish similar with `docker compose down` (with `--rmi local` if you want to nuke images). Feels unnecessary overcomplication here, tbh. It's an example app after all, it should be easy to setup and as easy to tear down without a trace.

examples/miniapps/faststream/src/producer.py
birthdaysgift: yes, for the sake of simplicity)

since we already have `fastapi-redis` example do you think this example needs DI container for redis as well?

docs/examples/faststream.rst
birthdaysgift: done

docs/examples/faststream.rst
birthdaysgift: done

examples/miniapps/faststream/Dockerfile.producer
birthdaysgift: done

examples/miniapps/faststream/README.rst
birthdaysgift: done

examples/miniapps/faststream/README.rst
birthdaysgift: it seems that I already committed `./run.sh` with `chmod +x` so I hope it works

removed this block from README.md

examples/miniapps/faststream/docker-compose.yml
birthdaysgift: Implemented via
```yaml
build:
  dockerfile: Dockerfile
```

for both `consumer` and `producer`

docs/examples/faststream.rst
birthdaysgift: done

examples/miniapps/faststream/run.sh
birthdaysgift: done

examples/miniapps/faststream/run.sh
birthdaysgift: it seems that I already committed ./run.sh with chmod +x so I hope it works

examples/miniapps/faststream/Dockerfile.consumer
ZipFile: I generally prefer src layout for clear code separation. In larger projects your root of the repo tends to be polluted with other stuff (e.g. django's `manage.py`, tons of tool configs, etc...).

For example app I think it would be better to stick to flat layout, for simplicity. I.e. remove `src/`.

examples/miniapps/faststream/src/producer.py
ZipFile: I'm not exactly familiar with how you usually separate publishers and subscribers in your faststream apps, but I have gut feeling that you want to reuse same configuration/container for both.

In particular, I think it would be nice to reuse the same broker instead of directly sending messages with raw redis client. For the sake of demonstration. https://faststream.ag2.ai/latest/redis/pubsub/publishing/#basic-redis-channel-publishing

docs/examples/faststream.rst
ZipFile: Did you push the changes?

examples/miniapps/faststream/src/producer.py
birthdaysgift: anyways, I added DI to `producer.py` in a separate commit, so you can decide whether it should be included or not

examples/miniapps/faststream/src/consumer.py
birthdaysgift: I tried to keep this example as small as possible to make key idea (adding `cast=False` to `Depends(...)`) to be clearly visible. In my opinion going full DI on this example will hide the important thing behind all other stuff that is not related to `FastDepends` integration.

But anyways I added full DI for `consumer.py` as a separate commit so you can decide whether it should be included or not.

So if you think that it's better to have full DI in this example I'll leave this decision up to you, I'm totally fine with both approaches.

docs/examples/faststream.rst
birthdaysgift: I started commenting along with adding changes so I can track what's done and what is not.

Now I committed everything so you can check

birthdaysgift: And thanks for the review btw :)

examples/miniapps/faststream/docker-compose.yml
birthdaysgift: you're right, fixed that

examples/miniapps/faststream/README.rst
```diff
@@ -0,0 +1,41 @@
+FastStream + Dependency Injector Example
+========================================
+
+This is a `FastStream <https://dummy.faststream.airt.ai/0.5/faststream/>`_ +
```
ZipFile:
```suggestion
This is a `FastStream <https://github.com/ag2ai/faststream>`_ +
```
Missed during first review phase.

examples/miniapps/faststream/faststream.rst
ZipFile, line 1: Looks like this is a duplicate of `docs/examples/faststream.rst`

examples/miniapps/faststream/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+name: faststream-example
+
+services:
+
+  redis:
+    image: redis
+
+  consumer:
+    build:
+      dockerfile: Dockerfile
```
ZipFile:
```suggestion
```

I'm pretty much sure this is default, no need to explicitly override to the same value.

examples/miniapps/faststream/src/__init__.py
ZipFile, line 1: Anyway, let's strip `src/` from example app structure. We do not want to overcomplicate things by introducing `pyproject.toml`, and without it `src` as an importable package is bad practice.

ZipFile (changes requested):
Few minor fixes and we're good to go.

examples/miniapps/faststream/docker-compose.yml
birthdaysgift: it doesn't work without `build` section, so I change it to be `build: .`

examples/miniapps/faststream/src/__init__.py
birthdaysgift: done

but I still cannot get why it is a bad practice for non-library projects :sweat_smile:

ZipFile: `src-layout` has strictly-defined semantics in Python world, many package managers (`poetry`, `uv`, etc...) expect `src/` folder to not be a package (otherwise they'll view it as `flat-layout` one). It just adds more confusion. And personally, `from src import ...` looks worse than `from myproject import ...`.

ZipFile (approved):
Thanks!

PR #963, opened 2026-05-11
allowing Configuration to use a pydantic class type instead of instance

warmagedon007:
This PR removes the error on loading and passing a `pydantic.BaseSettings` type when `from_pydantic` is called.

Instead, an instance will be directly initialized (with no kwargs) for use in the loading process.

The purpose of this and to allow declaration and import of a declarative container before all setting are initialized (via envvars, files, etc), especially in tests.

A little note.
kwargs that are used now in the `Configuration` are for passing kwargs for the pydantic `dict` or `model_dump`. I thought it will be too confusing to change it to be also kwargs for the pydantic model initialization.


ZipFile (changes requested):
That's a nice addition. But we need few more changes:
* Update typing info in `src/dependency_injector/providers.pyi` (+ add type assertions to  `tests/typing/configuration.py`)
* Update documentation (`docs/providers/configuration.rst`).

ZipFile (approved):

PR #965, opened 2026-05-18
Support wiring of Cython-compiled modules

keyz182:
## Note:
I used Claude Code heavily here. I don't see any policy on your stance on AI use, so if you're against it, I can drop the PR. 

That said, this works against our codebase when we run it through cythonize (previously we could only cythonize code that didn't have dependency injection). 

## Problem

`Container.wire()` silently skips functions and methods in user modules compiled to `.so` via Cython. `Depends(Provide[Container.x])` markers in those modules fall through unresolved as the literal `_Marker` sentinel, surfacing later as `AttributeError` from the handler body.

Failure shape (real-world FastAPI handler):

```
File "member_service/entrypoints/http/current_identity.py", line 21, in __call__
File "member_service/service/user.py", line 50, in get_user_account_by_email
AttributeError: __aexit__
```

The `uow` parameter received the `_Marker` sentinel because the wiring discovery pass never registered the cyfunction.

## Root cause

`wiring.py` gates on `inspect.isfunction` / `inspect.ismethod`, both of which return `False` for `cython_function_or_method` instances. Affected call sites in `wiring.py`:

- `wire()` module-member dispatch — `elif isfunction(member):` (skips compiled module-level functions)
- `_is_method()` — `ismethod(m) or isfunction(m)` (skips compiled class methods, including `__call__`)
- `unwire()` traversal — `if isfunction(member)` + `getmembers(member, isfunction)` (no-op cleanup for compiled targets)
- `_get_patched()` sync/async dispatch — `iscoroutinefunction` / `isasyncgenfunction` (Cython < 3 wraps coroutine cyfunctions with the sync patcher)

`_cwiring.pyx` is unaffected — it operates on the already-built `injections` dict and never inspects user functions. The blindness is entirely in the discovery pass.

## Safety analysis

Verified that dep-injector never mutates user functions:

- `_patch_fn` / `_patch_method` wrap via `functools.wraps` (writes go to the new pure-Python wrapper, not the cyfunction) and rebind the attribute on the parent module/class via `setattr`.
- `_fetch_reference_injections` only reads `inspect.signature(fn)`, which works on cyfunctions built with `embedsignature=True`.
- Zero writes to `fn.__defaults__`, `fn.__code__`, `fn.__globals__`, or `fn.__signature__`.

Cython's read-only restrictions on those slots are therefore irrelevant.

## Change

`src/dependency_injector/wiring.py`:

- New private helpers: `_is_cyfunction`, `_is_function_like`, `_iscoroutinefunction_compat`, `_isasyncgenfunction_compat`.
- Replace `isfunction(member)` at the four discovery gates with `_is_function_like(member)`.
- Replace `iscoroutinefunction(fn)` / `isasyncgenfunction(fn)` in `_get_patched` with the compat helpers, which fall back to `__code__.co_flags & CO_COROUTINE` / `CO_ASYNC_GENERATOR` so Cython < 3 async / async-gen cyfunctions are dispatched correctly.
- Detection scoped to `cython_function_or_method` only. `fused_cython_function` templates dispatch per call; wrapping the template would inject before type dispatch — intentionally excluded as potential follow-up work.

Pure-Python wiring behaviour is unchanged. `_is_function_like(obj) = isfunction(obj) or _is_cyfunction(obj)` only widens the accept set.

## Tests

`tests/unit/wiring/test_cython.py` — 15 tests covering:

- Cyfunction detection on compiled sync / async / async-gen functions; rejection on pure-Python.
- `_iscoroutinefunction_compat` / `_isasyncgenfunction_compat` on real compiled handlers AND via a synthetic `__code__.co_flags` callable to exercise the Cython < 3 fallback branches.
- Runtime injection for sync / async / async-gen / class `__call__` shapes.
- Provider override semantics across the compile boundary.
- Positive `_patched_registry` assertion that `unwire()` empties the injection bindings while preserving `reference_injections`.
- Descriptor-type pin on `HandlerClass.__call__` so a future Cython release that changes the class-attribute representation surfaces as a test failure rather than silent wiring drop-out.
- Module-level symbol-presence assertion so an empty / stale `.so` fails loud at collection time instead of producing a silent `0 collected` CI run.

`tests/unit/wiring/conftest.py` — session-start build of the `.pyx` fixture via `cythonize` + `build_ext` with bare module name + `inplace=0` + `build_lib` pointed at the fixture dir (the .so lands where the test importer expects, not at the repo root that a dotted-or-undotted name with `inplace=1` would produce on a clean checkout). Freshness check includes `EXT_SUFFIX` so a stray .so from a different ABI/Python version is treated as a miss. Build skips silently if Cython / setuptools / a C toolchain is unavailable.

`tests/unit/samples/wiringcython/cythonmodule.pyx` — fixture exercising the four wiring dispatch shapes with `binding=True`, `embedsignature=True`, `annotation_typing=False`.

`tox.ini` — `cython>=3,<4` added to `[testenv]` deps so the default `tox` invocation builds the fixture. Without it, `test_cython.py` silently skips.

`.gitignore` — fixture build artefacts (`*.c`, `*.so`, `_build/`).

`docs/wiring.rst` — new "Wiring of Cython-compiled modules" section documenting the required `cythonize` compiler directives (with a callout that `annotation_typing` defaults to `True` in Cython 3.x and must be flipped to `False` for FastAPI handlers using `param: str = Header(...)` patterns).

## Validation

| Check | Result |
|---|---|
| dep-injector own test suite + 15 new cython tests | **1378 passed, 0 failed, 2 skipped** (skips unrelated to this change) |
| Real-world FastAPI codebase compiled to 51 `.so` files, run against patched wiring.py | **304 passed, 0 cython-specific failures** — parity with the equivalent pure-Python image |
| Same codebase against unpatched wiring.py | 289 passed, **15 handler tests fail** with the unresolved-sentinel symptom |
| Behaviour parity vs unpatched wiring.py for pure-Python wiring | identical |

## Out of scope (follow-up)

- `_patch_fn` does `setattr(module, name, fn)` with no read-back verification. Read-only slot / descriptor edge cases could silently no-op. Pre-existing behaviour unchanged by this PR; flagged for a separate hardening pass.
- `fused_cython_function` template support. Would require deciding whether to wrap before type dispatch or per-resolved-instance.

## Reviewer notes

- `_is_cyfunction` uses `type(obj).__name__ == "cython_function_or_method"` rather than an `isinstance` check against a Cython-imported type, to avoid making Cython a runtime import requirement of dep-injector itself. The name has been stable across Cython 0.29 → 3.x.
- The `co_flags` fallback branches are unreachable from real Cython 3.x cyfunctions (the project pins Cython ≥ 3.1.4). They exist as defensive coverage for downstream users on older Cython. Synthetic tests exercise them.


tests/unit/wiring/conftest.py
ZipFile, line 1: This looks overly complicated. You can achieve similar just by running `cython -i filename.pyx` or by using `pyximport`. They already handle caching, cross platform and in general require less boilerplate code to use.

tests/unit/wiring/conftest.py
```diff
@@ -0,0 +1,111 @@
...
+_LOG = logging.getLogger(__name__)
+
+_FIXTURE_DIR = (
+    Path(__file__).resolve().parent.parent / "samples" / "wiringcython"
+)
+_PYX = _FIXTURE_DIR / "cythonmodule.pyx"
+
+
+def _ext_suffix() -> str:
+    return sysconfig.get_config_var("EXT_SUFFIX") or ".so"
```
ZipFile: Will it run on windows and mac?

tests/unit/wiring/conftest.py
```diff
@@ -0,0 +1,111 @@
...
+    if not _PYX.exists():
+        return False
+    if _fixture_already_built():
+        return True
+    try:
+        from Cython.Build import cythonize
+        from setuptools import Extension
+        from setuptools.command.build_ext import build_ext
+        from setuptools.dist import Distribution
+    except ImportError as exc:
```
ZipFile: Cython and setuptools are always installed when running tests. This is extra.

src/dependency_injector/wiring.py
```diff
@@ -127,6 +129,78 @@ def is_werkzeug_local_proxy(obj: Any) -> bool:
...
+    code = getattr(fn, "__code__", None)
+    if code is None:
+        return False
+    return bool(getattr(code, "co_flags", 0) & CO_COROUTINE)
+
+
+def _isasyncgenfunction_compat(fn: Any) -> bool:
+    """Async-generator check that also handles Cython-compiled ``async def`` w/ yield.
+
+    Symmetric to :func:`_iscoroutinefunction_compat`: Cython < 3.0
```
ZipFile: We've already migrated from older Cython versions. This is extra.

src/dependency_injector/wiring.py
```diff
@@ -127,6 +129,78 @@ def is_werkzeug_local_proxy(obj: Any) -> bool:
...
+    is still present in ``__code__.co_flags``. Without this helper, async-
+    gen cyfunctions would fall through to ``_get_sync_patched`` and break
+    at first ``await`` / ``async for``.
+    """
+    if isasyncgenfunction(fn):
+        return True
+    code = getattr(fn, "__code__", None)
+    if code is None:
+        return False
+    return bool(getattr(code, "co_flags", 0) & CO_ASYNC_GENERATOR)
```
ZipFile: `isasyncgenfunction` already does exactly the same check, but is more robust.
https://github.com/python/cpython/blob/6372d48a4d5c394a6638741d0c3c849a20161efb/Lib/inspect.py#L331-L337
https://github.com/python/cpython/blob/6372d48a4d5c394a6638741d0c3c849a20161efb/Lib/inspect.py#L285-L296

src/dependency_injector/wiring.py
```diff
@@ -127,6 +129,78 @@ def is_werkzeug_local_proxy(obj: Any) -> bool:
...
+    """Return True for pure-Python functions and Cython-compiled functions.
+
+    Wiring's discovery pass must accept both so that codebases compiled to
+    ``.so`` extensions (e.g. for source-protected container images) can be
+    wired transparently.
+    """
+    return isfunction(obj) or _is_cyfunction(obj)
+
+
+def _iscoroutinefunction_compat(fn: Any) -> bool:
```
ZipFile: See comments above.

tests/unit/wiring/test_cython.py
```diff
@@ -0,0 +1,239 @@
...
+# false-green CI mode that motivated the rest of this file. Fail loudly at
+# collection time instead.
+for _sym in ("sync_handler", "async_handler", "async_gen_handler", "HandlerClass"):
+    if not hasattr(cythonmodule, _sym):
+        raise AssertionError(
+            f"Cython fixture is missing required symbol: {_sym}. "
+            "The .so was importable but doesn't contain the expected handlers — "
+            "rebuild or clean stale artefacts under "
+            "tests/unit/samples/wiringcython/."
+        )
```
ZipFile:
```suggestion
```
Looks extra. We'll fail either way if sample module failed to build.

tests/unit/wiring/test_cython.py
```diff
@@ -0,0 +1,239 @@
...
+# with a callable carrying a forged ``__code__`` so the branch is covered.
+
+
+class _FakeCode:
+    def __init__(self, co_flags: int) -> None:
+        self.co_flags = co_flags
+
+
+class _FakeCallable:
+    """Inspection-only stand-in for an async cyfunction built on Cython < 3.
```
ZipFile: We do not support Cython<3. This is extra.

tests/unit/wiring/test_cython.py
```diff
@@ -0,0 +1,239 @@
...
+
+    assert not _is_cyfunction(py_fn)
+
+
+def test_is_function_like_accepts_both():
+    def py_fn():
+        pass
+
+    assert _is_function_like(py_fn)
+    assert _is_function_like(cythonmodule.sync_handler)
```
ZipFile: This can be merged into single parametrized test. Also `test_handler_class_call_is_function_like` fits in here too.

ZipFile (changes requested):
Good idea, but too much extra code and comments.

keyz182:
Thanks for the feedback, I'll dive in later today 👍

tests/unit/wiring/conftest.py
keyz182: Switched to pyximport

tests/unit/wiring/conftest.py
keyz182: pyximport should handle all platforms now

tests/unit/wiring/conftest.py
keyz182: Dropped.

src/dependency_injector/wiring.py
keyz182: Dropped

src/dependency_injector/wiring.py
keyz182: Simplified to use the stock functions.

tests/unit/wiring/test_cython.py
keyz182: Dropped

tests/unit/wiring/test_cython.py
keyz182: Dropped

tests/unit/wiring/test_cython.py
keyz182: Merged into parameterisde test.

keyz182:
Very much simplified now. Reduced to a ~15 line change to the library, the rest is tests.

docs/wiring.rst
```diff
@@ -658,6 +658,48 @@ or with a single container ``register_loader_containers(container)`` multiple ti
...
+           # import time.
+           "annotation_typing": False,
+       },
+   )
+
+No public API change in *Dependency Injector* is required to consume
+compiled modules — ``container.wire(packages=[my_package])`` /
+``container.wire(modules=[my_compiled_module])`` discover and patch
+cyfunctions alongside pure-Python functions in the same package tree.
+
```
ZipFile:
```suggestion
```
Looks like this should've been part of the PR message, not docs.

docs/wiring.rst
```diff
@@ -658,6 +658,48 @@ or with a single container ``register_loader_containers(container)`` multiple ti
...
+           "language_level": 3,
+           "binding": True,
+           "embedsignature": True,
+           # Keep annotation_typing=False for FastAPI handlers using
+           # `param: str = Header(...)` / `dep: Service = Depends(...)`:
+           # with annotation_typing=True (the Cython 3.x default!) Cython
+           # generates a C-level isinstance check against the default
+           # sentinel and raises `TypeError: Expected str, got Header` at
+           # import time.
+           "annotation_typing": False,
```
ZipFile:
```suggestion
```
Disabling it for entire modules does not look like a good idea.
Let's link to the [@cython.annotation_typing(False)](https://cython.readthedocs.io/en/latest/src/tutorial/pure.html#disabling-annotations) with short explanation why/where you should use it (separate sub-section in the doc).

tests/unit/samples/wiringcython/cythonmodule.pyx
```diff
@@ -0,0 +1,44 @@
...
+                              relies on this)
+  language_level=3          — pure Python 3 semantics
+
+Exercises every call shape the wiring discovery pass dispatches on:
+
+  - sync def at module level                    -> _get_sync_patched
+  - async def at module level                   -> _get_async_patched
+  - async def with yield (async generator)      -> _get_async_gen_patched
+  - class with async def __call__               -> _patch_method
+"""
```
ZipFile:
```suggestion
```
We've already described compiler options nuances in the docs, let's remove it from here.

tox.ini
```diff
@@ -18,6 +18,11 @@ deps=
     pydantic-settings
     werkzeug
     fast-depends
+    # Cython is required to build the .pyx fixture under
+    # tests/unit/samples/wiringcython/ that backs the wiring-vs-Cython
+    # regression coverage. Without it, tests/unit/wiring/test_cython.py
+    # skips silently.
```
ZipFile:
```suggestion
    # Cython is required to build tests/unit/samples/wiringcython/
```

ZipFile (changes requested):
Few minor changes to the docs and it is good to go.

keyz182:
Changes made :)

ZipFile:
> Changes made :)

Almost. Add few lines about `@cython.annotation_typing(False)` for fastapi views/deps.

docs/wiring.rst
```diff
@@ -688,6 +688,27 @@ extensions for a FastAPI / dependency-injector codebase:
        },
    )
 
+FastAPI views and dependencies that rely on parameter defaults as
```
keyz182, line 691: @ZipFile hows this?

keyz182:
Got a fix for the failing test - just testing locally.

keyz182:
Fix pushed

ZipFile (approved):
Thanks!

PR #972, opened 2026-07-03
Remove unused imports and variables (F401, F841)

Mubashir78:
Remove unused imports and local variables across type stubs and test files flagged by Ruff rules F401 and F841.


ZipFile (approved):
Thanks!

PR #973, opened 2026-07-03
Do not type DeclarativeContainer.__getattr__ as returning a Provider (#910)

### LINKED ISSUE #910: __getattr__ typing silences legitimate issues
https://github.com/ets-labs/python-dependency-injector/blob/master/src/dependency_injector/containers.pyi#L59

Having 
```
Container:
    
    def __getattr__(self, name: str) -> Provider: ...
```

in stubs silences cases, when container does not have particular attribute at-all. In my current codebase, I add snippet:

```
if typing.TYPE_CHECKING:
    from dependency_injector.containers import DeclarativeContainer as DC
    class DeclarativeContainer(DC):
        def __getattr__(self, name: str) -> typing.Never: ...
else:
```
to work around it.

Is it necessary to have this method typed like that?

uttam12331:
## Summary

Closes #910.

The container type stub declares

```python
class Container:
    ...
    def __getattr__(self, name: str) -> Provider[Any]: ...
```

on the **base** `Container`, so `DeclarativeContainer` inherits a catch-all `__getattr__`. That makes type checkers accept access to *any* attribute on a declarative container as a `Provider`, silently hiding typos and references to providers that were never declared:

```python
    service = providers.Factory(Service)

c = Container()
c.serivce  # typo — no type error today, inferred as Provider[Any]
```

As noted in the issue, this only makes sense for [`DynamicContainer`](https://python-dependency-injector.ets-labs.org/containers/dynamic.html), whose providers are added by name at runtime — not for [`DeclarativeContainer`](https://python-dependency-injector.ets-labs.org/containers/declarative.html), whose providers are declared statically.

## Change

Move the `__getattr__` stub from the base `Container` to `DynamicContainer`. Now:
- `DeclarativeContainer` exposes only its statically declared providers → accessing an undeclared attribute is flagged.
- `DynamicContainer` keeps resolving arbitrary attributes to `Provider[Any]`.

Verified with `mypy --strict`:

```python
class C(containers.DeclarativeContainer):
    provider = providers.Factory(int)

C().provider       # ok: Factory[int]
C().nonexistent    # error: "C" has no attribute "nonexistent"   <-- previously silent

containers.DynamicContainer().anything   # ok: Provider[Any]
```

## Tests

`mypy --strict tests/typing` passes (19 source files). Added Test 7 in `tests/typing/dynamic_container.py` asserting that dynamic attribute access still resolves to `providers.Provider[Any]`, so the `DynamicContainer` behavior stays covered. The existing `declarative_container.py` typing tests continue to pass because they access explicitly declared providers.


ZipFile (approved):
Thanks!

PR #979, opened 2026-08-07
Add .gitattributes for consistent line ending normalisation

chernojagne:
Add a `.gitattributes` file to normalise line endings and mark binary files consistently across platforms.

**What this adds**

- `* text=auto` — Git auto-detects text files and normalises line endings on checkout/commit, preventing CRLF/LF mix-ups between Windows and Unix contributors.
- Explicit `binary` markers for images, fonts, archives, compiled objects, executables, audio, video, Python bytecode/wheels, and lockfiles (`uv.lock`, `Pipfile.lock`, `poetry.lock`) — these are already binary and should never be diffed as text.
- `*.patch` / `*.diff` marked as `-text` to preserve byte sequences.

**Why it matters**

Without `.gitattributes`, cross-platform contributors can introduce invisible whitespace diffs and corrupt binary files during merge. This is a zero-controversy, one-file addition that follows the [gitattributes community standard](https://git-scm.com/docs/gitattributes).

---

This change was generated by [Autar](https://autar.ai) and reviewed by a human before submission.


ZipFile:
I don't think we have a problem with newlines and/or cross-platform contributions. I prefer [.editorconfig](https://editorconfig.org/) for this purpose (which we already have in our repo), since instructing git to fix newlines often ends up in unnecessary diff noise.

